PR #7, opened 2022-07-15
Add more e2e tests

achaikou:


tests/e2e/test_slice.py
```diff
@@ -34,9 +47,130 @@ def getslice(lineno, direction):
...
+    ("&comp=metadata"),
+    ("$comp=blobs&where=Name%20%3E%20%27C%27%20AND%20Name%20%3C%20%27D%27"),
+    ("$restype=container"),
+    ("&versionid=2020-01-01"),
+    ("&snapshot=2011-03-09T01:42:34.9360000Z"),
+    ("&timeout=1"),
+    ("&blackbox=surprise"),
+])
+@pytest.mark.xfail(reason="Figure out which are allowed. Some cause 500 because of params, some because of sas, some are 200")
+def test_unexpected_parameters_get(path, unexpected_params):
```
sveinung-r, line 167: This test seems a bit random. I'm not quite sure what we get from testing what status code we get from passing various query parameters to Azure?

sveinung-r (changes requested):

tests/e2e/test_slice.py
```diff
@@ -34,9 +47,130 @@ def getslice(lineno, direction):
...
+    }
+    """)
+    assert meta == expected_meta
+
+
+@pytest.mark.parametrize("method", [
+    ("get"),
+    ("post")
+])
+def test_crossline_slice(method):
```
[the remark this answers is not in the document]
achaikou: I am not sure (hence the comment at the end of the file).
1. From practical point of view, these tests (and actually slice tests for crossline/sample) are not needed. They all are expected to be tested on the unit level. These e2e actually need to make sure that yet untested code works correctly - which is go bindings. And from looking-at-the code we can see that paths don't actually differ in case of slice/inline/sample. It's just a parameter. So it doesn't have to be tested here.
2. From theoretical point of view, you want to test from e2e everything important that user might encounter. Because what if there are changes in the unit code...
So that's why I added inline/xline/samples slice tests - as these endpoints are kinda important.

So I am kinda unsure what should be double-tested here and for what unit tests are totally enough.

tests/e2e/test_slice.py
achaikou: Consider this a test of likes "test-driven programming" :D 

You and Erlend both added a comment to the code:
```
	/*
	 * This is super buggy as assumes that no other query-paramters are
	 * present.
	 */
```
	 
This is the test (placeholder) that responds to this comment. I looked at some current interesting params that are given further to azure blob and stored them here.
So whoever will try to solve this comment, has an idea what to look for. Do we remove everything or is snapshot something that should actually be passed? What about timeout? And so on.
When the decisions are made, everything that is of no interest should be deleted.

sveinung-r: We are now about to change the way we pass sas tokens, so this will not be needed.

achaikou: No, why?
Changes are just structural.
I still pass this weird parameter along with sas and it is still passed directly to blob. So we still have the same question as before.

sveinung-r: I think we should just trust that the sas parameter is in fact just the sas token. Otherwise we will have to make sure to keep the list of valid and invalid parameters up to date with new versions of Azure blob. Not sure this is worth it since you are not permitted to do any more damage through our system than your sas token permits you to do directly towards Azure blob storage anyway.

The chances of something accidentally being passed along is however smaller.

achaikou: 1. Refer to Erlend as it was him who first brought this problem up. I disagreed with him at first (because yes, that's Azure job), but he was adamant we need to check for these things, or our solution would be kinda insecure, hence that's what I put into tests placeholder.

2. It's not always about accidentally passing variables, but also maliciously on purpose. Erlend made a comment about "what if there is a parameter to delete delete all the resources" or something. My point was that if someone has possibility to do that, they won't need our API and can do it directly. However that made me check what parameters are present. Because if there was a delete parameter and someone for whatever reason passed it, they would be a really bad thing for our API at least from audit point of view.

3. There are apparently interesting parameters that are present and question is what do we want to do with them. As Erlend originally suggested - should we just remove everything that is not sas? Or should we just pass everything further? Openvds passes everything to azure, as it appears. So in case of some parameters we get 500 - because query is valid for azure but openvds doesn't expect the response it gets - but that's for input user unlikely wants to pass.
Other thing is that it has more interesting parameters, like snapshot and timeout. Are there any unexpected consequences of passing them? Could there appear new dangerous "delete-like" parameters in Azure in future ~(yes, in GET request, I know)~?

So to sum up: if everyone, Erlend included, thinks that there is no point in sanitizing what is passed to azure and no existing parameters cause any danger, I am totally fine with dropping these placeholder tests.

ErlendHaa: I've way less concerned about this now than I was previously. The main point for me is that you need a valid sas token to exploit anything (if there is anything to exploit that is). And as you say, if you have that you can just go directly to azure. W.r.t. to accidentally passing extra parameters I think that when sas is moved into the json payload, it's "sufficiently" hard to mess up. And we cannot forget to delete other query-parameters like before. 

All in all, I don't think this is an issue anymore. Even tho I made a big fuzz last time

PR #11, opened 2022-07-21
Make sas-token a part of payload in /slice & /slice/metadata

ErlendHaa:
The main work of this PR makes changes to how the slice endpoints accepts parameters. Additionally it does a bit of refactoring and misc work to facilitate these changes, as well as making the code more testable. Tests are left for the next PR

Endpoint /slice and /slice/metadata now expects the following payload:

```
{
    "direction": "inline",
    "lineno": 9961,
    "vds": "container/subpath",
    "sas": "sv=..."
}
```
Both endpoints still support GET and POST methods. The GET methods
expects the payload to be URLescaped and passed as query parameter
'query'. I.e. ?query=urlescape(payload.json). The POST methods expect
the json to be passed in the request body.

Examples
--------

Get:
```
curl -G --data-urlencode query@body.json "http://localhost:8080/slice/metadata"
```
Post:
```
curl -H "Content-Type: application/json" -d @body.json "http://localhost:8080/slice/metadata"
```
where the content of query.json equals the above example payload. This is similar to how GraphQL does POST and GET

### DIFF
--- a/api/query.go
+++ b/api/query.go
@@ -39,7 +39,7 @@ func (e *Endpoint) sliceMetadata(ctx *gin.Context, query SliceQuery) {
 		return
 	}
 
-	buffer, err := vds.SliceMetadata(conn.Url, conn.Credential, *query.Lineno, axis)
+	buffer, err := vds.SliceMetadata(*conn, *query.Lineno, axis)
 	if err != nil {
 		log.Println(err)
 		ctx.AbortWithStatus(http.StatusInternalServerError)
@@ -63,7 +63,7 @@ func (e *Endpoint) slice(ctx *gin.Context, query SliceQuery) {
 		return
 	}
 
-	buffer, err := vds.Slice(conn.Url, conn.Credential, *query.Lineno, axis)
+	buffer, err := vds.Slice(*conn, *query.Lineno, axis)
 	if err != nil {
 		log.Println(err)
 		ctx.AbortWithStatus(http.StatusInternalServerError)
--- a/api/query.go
+++ b/api/query.go
@@ -33,7 +33,6 @@ func (e *Endpoint) sliceMetadata(ctx *gin.Context, query SliceQuery) {
 	}
 
 	axis, err := vds.GetAxis(strings.ToLower(query.Direction))
-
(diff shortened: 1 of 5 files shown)


api/query.go
```diff
@@ -1,130 +1,136 @@
...
-	if err := ctx.ShouldBind(&query); err != nil {
-		ctx.AbortWithError(http.StatusBadRequest, err)
-		return
+	q, err := url.QueryUnescape(ctx.Query("query"))
+	if err != nil {
+		return nil, err
 	}
 
-	querystr := ctx.Request.URL.Query()
+	err = json.Unmarshal([]byte(q), &query)
```
achaikou, line 82: Passing something bad, like `{ "direction" :  "inline" } `causes 500. Why 500?
```
2022/07/21 12:13:22 [Recovery] 2022/07/21 - 12:13:22 panic recovered:
GET /slice?query={%20%22direction%22%20:%20%20%22inline%22%20} HTTP/1.1

Comment (ErlendHaa) [reply api/query.go:82]: Good catch. It's because I use json.Marshal for GET requests, which doesn't know about the binding tag in the SliceQuery struct. I.e. it's fine with missing fields and happily continues until we hit OpenVDS. which of course goes haywire on empty strings.

I cannot use gin.ShouldBind on Get request because it's urlencoded, so might need to manually verify that the fields are present

File: cmd/query/main.go [hunk 2]
```diff
@@ -55,10 +55,10 @@ func main() {
 
 	app := gin.Default()
 	app.GET("/", endpoint.Health)
-	app.GET("slice", endpoint.Slice)
-	app.POST("slice", endpoint.Slice)
-	app.GET("slice/metadata", endpoint.SliceMetadata)
-	app.POST("slice/metadata", endpoint.SliceMetadata)
+	app.GET( "slice", endpoint.SliceGet)
+	app.POST("slice", endpoint.SlicePost)
+	app.GET( "slice/metadata", endpoint.SliceMetadataGet)
```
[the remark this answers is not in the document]
ErlendHaa: On purpose

achaikou (approved):

PR #13, opened 2022-07-21
Add endpoint documentation with swagger

ErlendHaa:


cmd/query/main.go
```diff
@@ -46,6 +49,14 @@ func parseopts() opts {
 	return opts
 }
 
+// @title        VDS-slice API
```
achaikou, line 52: I suspect we are just never changing the name, are we?..
OK... I will have to learn to live with it then, even though I really find it misleading

api/query.go
```diff
@@ -122,6 +155,12 @@ func (e *Endpoint) SliceMetadataPost(ctx *gin.Context) {
 	e.sliceMetadata(ctx, *query)
 }
 
+// Slice godoc
+// @Summary  Fetch a single slice in any cube direction
+// @Param    query  query  string  True  "Urlencoded/escaped SliceQuery"
+// @Produce  application/octet-stream
+// @Success  200
+// @Router   /slice  [get]
```
achaikou, line 163: Why do I get "example" on post, but not on get? Can I have a "get" example?

api/query.go
```diff
@@ -13,13 +13,33 @@ import (
...
+	// i, j, k are zero-indexed and correspond to Inline, Crossline,
+	// Depth/Time/Sample, respectably.
+	//
+	// All options are case-insensitive.
+	Direction string `form:"direction" json:"direction" binding:"required"`
+
+	// Linenumber of the slice
+	Lineno *int `form:"lineno" json:"lineno" binding:"required"`
+
+	// A valid sas-token with read access to the container specified in Vds
```
achaikou, line 40: 1. Maybe decide if it is vds, VDS or Vds?
2. Can we test it again on Radix? I am very confused. Sometimes it seems to me like container sas is enough. But if I now try Radix, not just I need to specify Iline instead of Inline, but also container sas doesn't fit, account one is needed.

api/query.go
```diff
@@ -13,13 +13,33 @@ import (
...
+	// depth unit "m" or "ft", but not if the units are "ms", "s".
+	// Sample is valid when the depth is unitless.
+	//
+	// i, j, k are zero-indexed and correspond to Inline, Crossline,
+	// Depth/Time/Sample, respectably.
+	//
+	// All options are case-insensitive.
+	Direction string `form:"direction" json:"direction" binding:"required"`
+
+	// Linenumber of the slice
```
achaikou, line 37: Line( )number

internal/vds/vds.go
```diff
@@ -26,6 +26,36 @@ const (
...
+	Sample float64 `json:"sample"`
+
+	// Axis units
+	Unit string `json:"unit"`
+} // @name VdsAxis
+
+
+// @Description Slice metadata
+type Metadata struct {
+	// Data format. See https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/cppdoc/enum/OpenVDS_VolumeDataFormat.html
```
achaikou, line 50: Maybe we also need to tell that this is enum, so if you get 3 it means - count 4 from the top?

api/query.go
```diff
@@ -13,13 +13,33 @@ import (
...
+	Vds string `form:"vds" json:"vds" binding:"required"`
+
+	// Direction can be specified in two domains
+	// - Annotation. Valid options: Inline, Crossline and Depth/Time/Sample
+	// - Index. Valid options: i, j and k.
+	//
+	// Only one of Depth, Time and Sample is valid for a given VDS. Which one
+	// depends on the depth units. E.g. Depth is a valid option for a VDS with
+	// depth unit "m" or "ft", but not if the units are "ms", "s".
+	// Sample is valid when the depth is unitless.
```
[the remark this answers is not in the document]
ErlendHaa: I tend to agree. There is an issue about it too, as there are some inconsistencies at them moment. But, this is how the code behaves atm, hence the docs

api/query.go
ErlendHaa: Examples are deduced from `@Param`'s type. For post that is the SliceQuery, but for Get it's technically a string. Changing `string` to `SliceQuery` makes the docs incorrect/confussing. It's possible to add an examples value after the comment in `@Param`, but that again should reflect the expected type (url encoded string, not SliceQuery)

api/query.go
```diff
@@ -131,6 +150,13 @@ func (e *Endpoint) SliceGet(ctx *gin.Context) {
 	e.slice(ctx, *query)
 }
 
+// Slice godoc
+// @Summary  Fetch a single slice in any cube direction
+// @Param    body  body  SliceQuery  True  "Query Parameters"
+// @Accept   application/json
```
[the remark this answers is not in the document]
ErlendHaa: Yes true, but I ment to remove that support when I moved sas token into the payload. It doesn't work as nicely anymore. The reason for the support in the first place was how gin parses query parameters (see [1]) and that I used too liberal gin binding functions, which. yield both query-parameter and form-data support.

I don't think we should start out with supporting too many options, so gonna remove form altogether.

cmd/query/main.go
```diff
@@ -6,7 +6,10 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/pborman/getopt/v2"
+	"github.com/swaggo/gin-swagger"
+	"github.com/swaggo/files"
 
+	_ "github.com/equinor/vds-slice/docs"
```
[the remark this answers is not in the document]
ErlendHaa: I don't, hence the underscore. But it needs to be imported for swagger to pick up the generated files in `/docs`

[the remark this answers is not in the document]
ErlendHaa: I was hoping that we didn't need to check in the autogenerated swagger files (`/docs/*`), but that doesn't seam to be possible. Which is super annoying, as we need to keep running `swag init` every time we change some source docs

api/query.go
achaikou: Something seems off with sas tokens or who knows what, so I will test it later, but in theory container should be enough.

api/query.go
ErlendHaa: That's not what it expects, the value of query is url encoded, so it looks like:
`query=direction%3D%22inline%22%0Alineno%3D%229961%22%0Avds%3D%22volve%2F32_RLE%22%0Asas%3D%22s...`, which is not a nice example

cmd/query/main.go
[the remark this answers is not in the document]
ErlendHaa: What I don't like is that if we check it in, it easily gets out of sync. Let's say we check in docs/docs.go now. In my next pr I add a new endpoint with docstrings, but I forget to run `swag init`. Now `docs/docs.go` is out of sync

api/query.go
```diff
@@ -16,7 +16,7 @@ import (
 // Query for slice endpoints
 // @Description Query payload for slice endpoints /slice and /slice/metadata.
 type SliceQuery struct {
-	// The blob path to a vds on form: container/subpath
+	// The blob path to a vds off form: container/subpath
```
achaikou, line 19: The switch was on, now you turned it off? :smile: 

Maybe `in form` then would be better than `of form`?

cmd/query/main.go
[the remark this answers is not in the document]
achaikou: But without that one who builds the code and has no interest in building docs is forced to install swag.

If it gets out of sync then it would not be a problem on server as it would be regenerated every time so users won't suffer.

We also can do something like create an empty docs/go file so that "install" doesn't complain, but with no actual content in it, and commit it? Not sure what will happen if we then add it to gitignore - will the local changes to it be ignored?
Or when testing we run swag init with output types which are not "go" and thus this file never changes?

ErlendHaa:
Closes #6

achaikou (commented):
OK, lets leave fetch docs for later

api/query.go
```diff
@@ -19,7 +19,7 @@ type FenceQuery struct {
 }
 
 // Query for slice endpoints
-// @Description Query payload for slice endpoints /slice and /slice/metadata.
+// @Description Query payload for slice endpoints /slice.
```
achaikou, line 22: It is just one endpoint now :smiling_imp:

docs/slice.md
```diff
@@ -0,0 +1,22 @@
+# Fetch a slice form a VDS
+
+Fetch a slice in any direction. Support 0-indexing of lines
```
achaikou, line 3: Support(s) maybe?

docs/slice.md
```diff
@@ -0,0 +1,22 @@
...
+and index-by-annotation such as inline and crossline numbers
+and depth intervals. See model SliceData for more info
+on request parameters.
+
+## Response
+On success (200) the response consists of two parts, metadata
+and data
```
achaikou, line 10: dot!

achaikou (approved):

PR #18, opened 2022-08-04
Add go unit tests to github actions

achaikou:
As we still build openvds from scratch, it takes a long time and might
not be very efficient to run on every PR.


ErlendHaa (approved):
LGTM! (maybe add "to gh" in the commit msg header, otherwise it's a tiny bit misleading)

PR #19, opened 2022-08-05
Simplify docker image

achaikou:
Not run on Radix yet, but hopefully will run and not report so many issues as final size is way smaller.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,15 @@
-# project is for the moment server only (no client in the same repo), so it deserves to be in the root dir
-
-FROM golang:1.18 as builder
-
-ENV DEBIAN_FRONTEND="noninteractive"
-RUN apt-get update
-RUN apt-get install -y \
-    build-essential \
-    git curl \
-    libboost-dev \
-    libcurl4-openssl-dev \
-    libuv1-dev \
-    libboost-all-dev \
+FROM golang:1.18-alpine as openvds
+RUN apk --no-cache add \
+    git \
+    g++ \
+    gcc \
+    make \
+    cmake \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,65 +1,76 @@
...
+    -DBUILD_DOCS=OFF \
+    -DDISABLE_AWS_IOMANAGER=ON \
+    -DDISABLE_AZURESDKFORCPP_IOMANAGER=OFF \
+    -DDISABLE_GCP_IOMANAGER=ON \
+    -DDISABLE_DMS_IOMANAGER=OFF \
+    -DDISABLE_STRICT_WARNINGS=OFF
 RUN cmake --build build   --config Release  --target install  -j 8 --verbose
 
 
+FROM openvds as vdsslice
```
ErlendHaa, line 35: Call it builder and explicitly builder? Then you differentiate build errors from test/install errors

Dockerfile
```diff
@@ -1,65 +1,76 @@
...
+FROM vdsslice as tester
 ARG CGO_CPPFLAGS="-I/open-vds/Dist/OpenVDS/include"
 ARG CGO_LDFLAGS="-L/open-vds/Dist/OpenVDS/lib"
+ARG LD_LIBRARY_PATH=/open-vds/Dist/OpenVDS/lib:$LD_LIBRARY_PATH
+ARG OPENVDS_AZURESDKFORCPP=1
+RUN go test -race ./...
 
-COPY . .
+
+FROM vdsslice as builder
```
ErlendHaa, line 50: installer?

ErlendHaa (approved):

PR #21, opened 2022-08-10
Tests update

achaikou:


tests/e2e/test_requests.py
```diff
@@ -0,0 +1,195 @@
...
+
+    expected_meta = json.loads("""
+    {
+        "axis": [
+            {"annotation": "Crossline", "max": 11.0, "min": 10.0, "samples" : 2, "unit": "unitless"},
+            {"annotation": "Sample", "max": 16.0, "min": 4.0, "samples" : 4, "unit": "ms"}
+        ],
+        "format": 3
+    }
+    """)
```
ErlendHaa, line 37: You could also use a regular dict

tests/e2e/test_requests.py
```diff
@@ -0,0 +1,195 @@
...
+VDS = "wellknown/well_known_default"
+
+
+@pytest.mark.parametrize("method", [
+    ("get"),
+    ("post")
+])
+def test_slice(method):
+    meta, slice = request_slice(method, 5, 'inline')
+    expected = np.array([[116, 117, 118, 119],
```
ErlendHaa, line 25: These tests have 3 components, 1) defining some expected value 2) calling the functions we are testing 3) asserting 1 and 2.

Can we reflow the test in that way? Now it's all a mixed together

tests/e2e/test_requests.py
```diff
@@ -0,0 +1,195 @@
...
+    ),
+    (
+        "fence",
+        {"vds": f'{CONTAINER}/{VDS}', "coordinate_system": "ij",
+            "coordinates": [[1, 2, 3]], },
+        http.HTTPStatus.INTERNAL_SERVER_ERROR,
+        "expected [x y] pair"
+    ),
+])
+def test_errors_are_forwarded(path, query, error_code, error):
```
ErlendHaa, line 116: Test error messages? are_forwarded refers to internal details for the server.

api/query_test.go
```diff
@@ -0,0 +1,330 @@
...
+				Max        float32 `json:"max"        binding:"required"`
+				Min        float32 `json:"min"        binding:"required"`
+				Samples    int     `json:"samples"    binding:"required"`
+				Unit       string  `json:"unit"       binding:"required"`
+			}
+
+			type SliceMetadata struct {
+				Axis   []SliceAxis `json:"axis"   binding:"required"`
+				Format int         `json:"format" binding:"required"`
+			}
```
ErlendHaa, line 184: Once #13 is merged, these will be available in the vds package

api/query_test.go
```diff
@@ -0,0 +1,330 @@
...
+		if testcase.expectedError == "" {
+			parts := readMultipartData(t, w)
+			if len(parts) != 2 {
+				msg := "Got %d parts in reply; want it to always contain 2 in case '%s'"
+				t.Errorf(
+					msg,
+					len(parts),
+					testcase.name,
+				)
+			}
```
ErlendHaa, line 171: The fact that you have such a big conditional for a couple of test cases suggest to me that maybe you want 2 tests, 1 for the successful requests and one for the failing ones

api/query_test.go
```diff
@@ -0,0 +1,330 @@
...
+	testcases := []SliceTest{
+		{
+			name:   "Valid GET Query",
+			method: http.MethodGet,
+			slice: TestSliceQuery{
+				Vds:       well_known,
+				Direction: "i",
+				Lineno:    0, //side-effect assurance that 0 is accepted
+				Sas:       "n/a",
+			},
```
ErlendHaa, line 68: Would it make sense to have a http.Request field here in stead?
```
request: makeRequest('get', '/slice', well_known, SliceQuery{...} )
```

i.e. something similar to the prepare function. You can even capture both slice and fence queries  (with `interface{}`)  to avoid have 2 separate functions like you do now.

api/query_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+			inlineAxis := SliceAxis{
+				Annotation: "Inline", Max: 5.0, Min: 1.0, Samples: 3, Unit: "unitless",
+			}
+			crosslineAxis := SliceAxis{
+				Annotation: "Crossline", Max: 11.0, Min: 10.0, Samples: 2, Unit: "unitless",
+			}
+			sampleAxis := SliceAxis{
+				Annotation: "Sample", Max: 16.0, Min: 4.0, Samples: 4, Unit: "ms",
+			}
```
ErlendHaa, line 194: Testing stuff like this should be tested in `vds_test.go` instead I think. Here we should only test the correspondence between metadata and data. I.e. the data length thing at the bottom here

api/query_test.go
```diff
@@ -0,0 +1,330 @@
...
+		} else {
+			assertError(w, t, struct {
+				name  string
+				error string
+			}{testcase.name, testcase.expectedError})
+		}
+	}
+}
+
+func assertError(w *httptest.ResponseRecorder, t *testing.T,
```
ErlendHaa, line 398: isn't it a bit cleaner to have this function return an error, which is asserted in the test? Then you don't need to pass around the test object (`t`).

api/query_test.go
```diff
@@ -0,0 +1,330 @@
...
+					msg,
+					len(parts[2]),
+					expectedDataLength,
+					testcase.name,
+				)
+			}
+		} else {
+			assertError(w, t, struct {
+				name  string
+				error string
```
ErlendHaa, line 262: This struct seams to contribute nothing but noise 😄

api/query_test.go
achaikou: `vds_test.go` should properly test that, if there are some unexpected/different cases or whatever.
 
But you need to happy-test it once here too (to make sure that metadata is as expected, no json stuff was lost in query.go get/parse or in the middleware), as we consider these our most outer tests run on every change.

I like to indirectly "randomize" parameters a bit (do not use the same value everywhere), but I again forgot that with removal of form I removed half of the cases where this randomization was used.
Removed unused to make stuff below a bit simpler.

api/query_test.go
achaikou: :woman_facepalming: 
(guess who managed to forget that nothing prevents her from passing values here as parameters)

api/query_test.go
```diff
@@ -395,24 +392,35 @@ func TestFenceHTTPResponse(t *testing.T) {
 	}
 }
 
-func assertError(w *httptest.ResponseRecorder, t *testing.T,
-	testcase struct {
-		name  string
-		error string
-	}) {
-	testErrorInfo := &TestErrorInfo{}
+func setupMock(r *gin.Engine) {
```
ErlendHaa, line 408: setupMockServer? or setupTestServer, as it's not mocking, its running production code in a test environment

api/query_test.go
ErlendHaa: Then be consistent at least, and move them out of the test-scope. Then the test becomes a bit denser too

api/query_test.go
ErlendHaa: I'm arguing that `expectedDataLength := expectedMetadata.Axis[0].Sample` is probably giving us enough assurance that the metadata is correct. After all, nothing really happens with the metadata in the http layer, other than writing a buffer. 

But if you insists all of this setup out-benefits the noise it creates, then I leave it alone 😄

ErlendHaa (approved):

PR #22, opened 2022-08-12
Request parsing cleanup

ErlendHaa:


api/query.go
```diff
@@ -57,32 +57,25 @@ func (e *Endpoint) slice(ctx *gin.Context, query SliceQuery) {
...
-
-	err := json.Unmarshal([]byte(ctx.Query("query")), &query)
-	if err != nil {
-		return nil, err
-	}
-
-	if err = binding.Validator.ValidateStruct(&query); err != nil {
-		return nil, err
+func parseGetRequest(ctx *gin.Context, v interface{}) error {
+	if err := json.Unmarshal([]byte(ctx.Query("query")), v); err != nil {
```
achaikou, line 61: I think I found it more readable when err assignment wasn't inside if.
Now the actual action - unmarshalling - seems more like a side-effect and err - the actual result value.

Same to other 2 uses.

api/query.go
```diff
@@ -57,32 +57,25 @@ func (e *Endpoint) slice(ctx *gin.Context, query SliceQuery) {
...
+		return err
 	}
 
-	return &query, nil
+	return binding.Validator.ValidateStruct(v);
```
achaikou, line 65: We also might want to discuss and agree what we want to do with formatters.
Default golang formatter (no idea which one it is), allows formatting only the full document, not the selection. So that means that if use formatter later on unformatted file, many lines will appear changed and it might create difficulties on review.

In this case btw it removes semicolon :D

achaikou (approved):

PR #33, opened 2022-08-18
Fence request names

### LINKED ISSUE #28: Rename struct fields / json tags in FenceRequest such that they match

ErlendHaa:
Closes #28

### DIFF
--- a/api/query.go
+++ b/api/query.go
@@ -32,10 +32,10 @@ type FenceRequest struct {
 	// cdp  : Coordinates are given as cdpx/cdpy pairs. In the original SEGY
 	//        this would correspond to the cdpx and cdpy fields in the
 	//        trace-headers after applying the scaling factor.
-	CoordinateSystem string `json:"coordinate_system" binding:"required"`
+	CoordinateSystem string `json:"coordinateSystem" binding:"required"`
 
 	// A list of (x, y) points in the coordinate system specified in
-	// coordinate_system.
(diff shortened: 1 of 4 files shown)


api/query.go
```diff
@@ -32,11 +32,11 @@ type FenceRequest struct {
 	// cdp  : Coordinates are given as cdpx/cdpy pairs. In the original SEGY
 	//        this would correspond to the cdpx and cdpy fields in the
 	//        trace-headers after applying the scaling factor.
-	CoordinateSystem string `json:"coordinate_system" binding:"required"`
+	CoordinateSystem string `json:"coordinateSystem" binding:"required"`
 
 	// A list of (x, y) points in the coordinate system specified in
 	// coordinate_system.
```
achaikou, line 38: Update name here

api/query_test.go
```diff
@@ -367,7 +367,7 @@ func TestFenceErrorHTTPResponse(t *testing.T) {
 			fenceQuery: "{\"vds\":\"" + well_known +
 				"\", \"coordinate_system\":\"ilxl\", \"sas\": \"n/a\"}",
```
achaikou, line 368: It is not really important (in theory all parameters can be removed altogether, as we just check for non-existence of one particular), but maybe update coordinate_system here and in neighbor test?

achaikou (approved):

PR #36, opened 2022-08-29
Valgrind

achaikou:
I am so, so, so lost.
Every single configuration makes it behave totally differently.

- Setup didn't cause any leaks on my local machine.
- Same setup on alpine run with local docker caused some errors.
- So I created an ubuntu setup which didn't cause failures with local docker.
- But then same ubuntu setup fails with new errors when run through github actions...

In the end I put some more tests in failing category and somehow everything else runs without a leak.....

Maybe we really should drop it all already now and leave memory/server restart management to Radix.
I am clearly out of my depth.


ErlendHaa:
> Every single configuration makes it behave totally differently.
 
If we don't know what causes the different behaviour the value of these tests drop quite a bit imo. Yes, we can rig it with a setup that doesn't fail on CI, but what value does it give? Our production env is different from the CI env. I think the cost of this uncertainty outweighs the benefits, assuming it's caused by third party library configurations (i.e. openvds)

achaikou:
Weird stuff seems to be caused by combination of other libraries and systems, not openvds on its own.

Now it is run exactly as it is in production - in alpine and with new azure sdk. So the "leaking" tests I suspect are leaking in production as well.

I agree with you that dropping it altogether now starts being even more tempting, but would love to hear Sveinung's opinion as well.

PR #38, opened 2022-09-01
Metadata tests

achaikou:
Again not sure if that is useful in any way, but probably won't hurt.

### DIFF
--- a/api/query_test.go
+++ b/api/query_test.go
@@ -3,6 +3,7 @@ package api
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"mime"
 	"mime/multipart"
@@ -14,6 +15,7 @@ import (
 	"testing"
 
 	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/require"
 )
 
 const well_known = "../testdata/wellknown/well_known_default.vds"
@@ -36,6 +38,15 @@ type fenceTest struct {
 	expectedError  string
 }
 
+type metadataTest struct {
+	name           string
+	method         string
+	metadata       testMetadataQuery
+	metadataQuery  string
+	expectedStatus int
+	expectedError  string
+}
+
 // define own help types to assure separation between production and test code
 type testErrorResponse struct {
 	Error string `json:"error" binding:"required"`
@@ -53,6 +64,11 @@ type testFenceQuery struct {
 	CoordinateSystem string      `json:"coordinateSystem"`
 	Coordinates      [][]float32 `json:"coordinates"`
 	Sas              string      `json:"sas"`
+}
+
+type testMetadataQuery struct {
+	Vds string `json:"vds"`
+	Sas string `json:"sas"`
 }
 
 type testSliceAxis struct {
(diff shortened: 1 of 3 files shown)


api/query_test.go
```diff
@@ -405,6 +421,127 @@ func TestFenceErrorHTTPResponse(t *testing.T) {
...
+				{"annotation": "Inline", "max": 5.0, "min": 1.0, "samples" : 3, "unit": "unitless"},
+				{"annotation": "Crossline", "max": 11.0, "min": 10.0, "samples" : 2, "unit": "unitless"},
+				{"annotation": "Sample", "max": 16.0, "min": 4.0, "samples" : 4, "unit": "ms"}
+			],
+			"format": 3
+		}`
+
+		if metadata != expectedMetadata {
+			msg := "Metadata not equal in case '%s'"
+			require.JSONEq(t, expectedMetadata, metadata, fmt.Sprintf(msg, testcase.name))
```
ErlendHaa, line 476: Is it worth a dependancy just to test the stringified response ? As opposed to just Marshall it into a struct? I know you prefer string comparison over struct (although I don't buy it 😄 ), but you already to that in the performance tests. So is it any point in doing it here too?

api/query_test.go
```diff
@@ -405,6 +421,127 @@ func TestFenceErrorHTTPResponse(t *testing.T) {
...
+			expectedStatus: http.StatusBadRequest,
+			expectedError:  "Error:Field validation for 'Vds'",
+		},
+		{
+			name:   "Query which passed all input checks but still should fail",
+			method: http.MethodPost,
+			metadata: testMetadataQuery{
+				Vds: "unknown",
+				Sas: "n/a",
+			},
```
ErlendHaa, line 517: Why do you use `metadataQuery` in some cases and `metadata` in others?

api/query_test.go
```diff
@@ -36,6 +38,15 @@ type fenceTest struct {
 	expectedError  string
 }
 
+type metadataTest struct {
+	name           string
+	method         string
+	metadata       testMetadataQuery
+	metadataQuery  string
```
ErlendHaa, line 45: I really dislike that you have this "either/or" design. It introduce too many questions and ambiguity. What's the rules for when to use witch? What purpose do they serve (one vs the other)? What happens if I specify both?

api/query_test.go
achaikou: The point was that I really didn't want to introduce again the same amount of code as for slice metadata - I remember you complained last time :smile: and I myself have spent some time considering to just copy this thing here again, but it didn't sit right with me. Strings comparison looked like an easy option.

So I decided that additional dependency might be a good idea. If I understand correctly, it won't be included in final build image.

Let me know if you for some reason still prefer it to be the other way around.

api/query_test.go
achaikou: Because metadata is nice and simple struct, visible values, and metadata query is ugly inline json.

I want to test ugly bad inline json in each call because in the code itself we have duplicated code for that (and I don't think it can be unified). For this I can't use a struct, because unmarshalling won't create invalid json.

For missing parameters tests I now can actually use the struct because we uncoupled the test structs and code structs. Tests structs no longer have "required" binding, so it is possible now! Didn't notice before. Would look nicer to me, not sure about you.

api/query_test.go
achaikou: Purpose answered in another comment.
idea was to use struct where possible for nicer readability and jsonQuery where desired json can't be represented as a struct.

Idea is that Query takes precedence over struct as it is more "final" form.

But let me know, if you want to combine everything to json, without structs at all, it is doable. Code would probably be less readable, but it might bring more additional benefits as getting rid of all the testQuery structs you despise so!

api/query_test.go
ErlendHaa: You could use the built-in reflect package to compare structs: 

https://github.com/equinor/vds-slice/blob/a0574cda9e075303d472ce5d2e94a286c06ba36c/internal/vds/vds_test.go#L65

api/query_test.go
ErlendHaa: > I want to test ugly bad inline json in each call because in the code itself we have duplicated code for that (and I don't think it can be unified). For this I can't use a struct, because unmarshalling won't create invalid json.

I think that the class of "invalid json" test cases should be it's own test: TestInvalidJsonBody 

>For missing parameters tests I now can actually use the struct because we uncoupled the test structs and code structs.

Just to clarify, if you construct a struct without a particular field, that field will be default constructed to its type's default value. Marshalling/Unmarshalling works fine, regardless of the `required` tag. Which means you can construct the test case and fire of the request just fine. The struct validator in our request handlers checks that the unmarshalled struct have no field with default values.

TL;DR you can even use the "code" struct in the test case when testing missing fields. This will work:

```
		{
			name:   "Query which passed all input checks but still should fail",
			method: http.MethodPost,
			fence: testFenceQuery{
				Vds:              "unknown",
				CoordinateSystem: "ilxl",
				Coordinates:      [][]float32{{3, 12}, {2, 10}},
				Sas:              "n/a",
			},
			expectedStatus: http.StatusInternalServerError,
			expectedError:  "Could not open VDS",
		},
```

>  Would look nicer to me, not sure about you.

I would definitely prefer to keep json-strings to a minimum wherever possible

api/query_test.go
ErlendHaa: > idea was to use struct where possible for nicer readability and jsonQuery where desired json can't be represented as a struct.

I would solve that to have different tests for "invalid / incomplete json" cases and other tests.

> But let me know, if you want to combine everything to json, without structs at all, it is doable. Code would probably be less readable

No thank you 😄 
> but it might bring more additional benefits as getting rid of all the testQuery structs you despise so!
See my other comment, you don't _need_ them in the first place

ErlendHaa (approved):

PR #39, opened 2022-09-05
Add gzip support to endpoints

ErlendHaa:
This commits add support for clients to ask for their response to be
gzip'd by specifying 'gzip' as part of the 'Accept-Encoding' header.

At the moment we priorities compression performance over compression
ratio. Compressing to aggressivly will have a negative impact on
cloud-to-cloud. This might change if we do more benchmarking.

The current compression ratio is about 2 for a farily dense binary
fence.

### DIFF
--- a/cmd/query/main.go
+++ b/cmd/query/main.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pborman/getopt/v2"
 	"github.com/swaggo/gin-swagger"
 	"github.com/swaggo/files"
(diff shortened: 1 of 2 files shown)


achaikou:
Why does majority of E2E tests fail with no changes?
Maybe add test for gzipped scenario to show how it is supposed to be handled? When I specify `application/gzip` header, nothing happens in the response, so I am doing something wrong?

achaikou (approved):
Looks like should be working as intended :+1:

PR #41, opened 2022-09-07
Add fence metadata

achaikou:


api/query.go
```diff
@@ -244,7 +251,7 @@ func (e *Endpoint) FenceGet(ctx *gin.Context) {
 // @Param    body  body  FenceRequest  True  "Request Parameters"
 // @Accept   application/json
 // @Produce  multipart/mixed
-// @Success  200
+// @Success  200 {object} vds.FenceMetadata
```
ErlendHaa, line 254: Maybe add a comment that FenceMetadata only applies to the first part in the response? See slice equivalent

api/query_test.go
```diff
@@ -332,14 +333,16 @@ func TestFenceHappyHTTPResponse(t *testing.T) {
...
-		if len(parts[0]) != 0 {
-			msg := "Got %d bytes in metadata reply; want it to always be 0 in case '%s'"
-			t.Errorf(
-				msg,
-				len(parts[1]),
-				testcase.name,
-			)
+		metadata := string(parts[0])
+		expectedMetadata := `{
+			"shape": [` + strconv.Itoa(len(testcase.fence.Coordinates)) + `, 4]
```
ErlendHaa, line 338: You really hate `fmt.Sprintf` 😄

internal/vds/vds.cpp
```diff
@@ -518,6 +518,42 @@ struct vdsbuffer fetch_fence(
...
+
+    auto access = OpenVDS::GetAccessManager(handle);
+    auto const *layout = access.GetVolumeDataLayout();
+
+    if (layout->GetDimensionality() != 3) {
+        throw std::runtime_error(
+            "Unsupported VDS, expected 3 dimensions, got " +
+            std::to_string(layout->GetDimensionality())
+        );
+    }
```
ErlendHaa, line 541: This should be extracted into a helper, as it should really be checked on every request. But it can be a TODO for later, as the same applies for other parts of these functions

internal/vds/vds.go
```diff
@@ -55,6 +55,12 @@ type Metadata struct {
 	Axis []*Axis `json:"axis"`
 } // @name Metadata
 
+// @Description Fence metadata
+type FenceMetadata struct {
+	// Describes the shape of returned data array.
```
ErlendHaa, line 60: "Shape of the returned data fence"? "Describes the" adds nothing, and data array is needlessly unprecise 😄

tests/e2e/test_requests.py
```diff
@@ -221,13 +230,16 @@ def request_fence(method, coordinates, coordinate_system):
 
     multipart_data = decoder.MultipartDecoder.from_response(rdata)
     assert len(multipart_data.parts) == 2
-    metadata = multipart_data.parts[0].content
+    metadata = json.loads(multipart_data.parts[0].content)
     data = multipart_data.parts[1].content
 
-    assert len(metadata) == 0
+    shape = (
```
ErlendHaa, line 236: `shape = tuple(metadata["shape"])`

Or maybe numpy even accept the list directly? Without the tuple cast?

ErlendHaa (commented):
LGTM!

api/query_test.go
achaikou: I think I can't use it in this case?
This is is a raw string literal (so that I can have all json quotes and so without problems and nice multiline).
I don't think it is possible for fmt to play nicely with it and it defeats purpose of raw string literal.

internal/vds/vds.cpp
achaikou: Yep, I think we agreed that this refactoring goes separately.

ErlendHaa (approved):

PR #46, opened 2022-09-13
Update swagger docs

achaikou:


api/query.go
```diff
@@ -240,6 +248,8 @@ func (e *Endpoint) SlicePost(ctx *gin.Context) {
 // @Accept   application/json
 // @Produce  multipart/mixed
 // @Success  200 {object} vds.FenceMetadata "(Example below only for metadata part)"
+// @Failure  400 {object} ErrorResponse "Request is invalid"
+// @Failure  500 {object} ErrorResponse "openvds failed to process the request"
```
ErlendHaa, line 252: Nothing is ever perfect when it comes to docs. Both of these render with the example value from ErrorResponse,  which looks kinda weirds, as the description ("Request is invalid") is not visible in the same view. Can we inline a custom error on 400?

api/query_test.go
```diff
@@ -555,18 +555,16 @@ func MakeFileConnection() vds.ConnectionMaker {
...
+	r.POST("/fence", ErrorHandler, endpoint.FencePost)
+	r.GET("/fence", ErrorHandler, endpoint.FenceGet)
 
-	r.POST("/fence", endpoint.FencePost)
-	r.GET("/fence", endpoint.FenceGet)
-
-	r.POST("/metadata", endpoint.MetadataPost)
-	r.GET("/metadata", endpoint.MetadataGet)
+	r.POST("/metadata", ErrorHandler, endpoint.MetadataPost)
+	r.GET("/metadata", ErrorHandler, endpoint.MetadataGet)
```
ErlendHaa, line 567: Maybe we should make a group and apply the error handler to that group? https://github.com/gin-gonic/gin#grouping-routes

achaikou: Thought to do so originally, but as far as I can tell, the group must have a name which is then a part of how you access the endpoints. So if I do group `e.Group("/stuff")`, I have to access all the endpoints added to it as `localhost:8080/stuff/fence`
Decided that we probably don't want that.
Now I see it [should be possible](https://github.com/go-ggz/ggz/blob/46af19ffdbe9b3d5316cfd4131eb3e0849297f17/pkg/router/routes/routes.go#L126-L133) to create 3 groups - one for metadata, one for fence, one for slice, and just combine post and get requests in them. Would you prefer that?

ErlendHaa: You're right, we don't want that :D

api/query.go
ErlendHaa: Yeah, I think a more generic error is better. Maybe even `{ "error": "message" }`

PR #47, opened 2022-09-15
Improve validations

achaikou:


internal/vds/vds.cpp
```diff
@@ -319,18 +319,26 @@ void json_write_axis(
...
+    return handle;
+}
+
+struct vdsbuffer fetch_slice(
+    std::string url,
+    std::string credentials,
+    axis ax,
+    int lineno
+) {
+    OpenVDS::ScopedVDSHandle handle = open_vds(url, credentials);
```
ErlendHaa, line 350: No auto?

internal/vds/vds.h
```diff
@@ -23,6 +23,12 @@ enum axis {
     SAMPLE    = 7,
 };
 
+enum coordinate_system {
+    IJ   = 0,
+    ILXL = 1,
+    CDP  = 2,
+};
+
```
ErlendHaa, line 31: There is a coordinate system enum in cpp, which is used for slice. We should unify them. And also agree on coordinate system names. It's a different commit for sure, but it would make sense to genereralizing the existing one, and then make use of it in this commit

internal/vds/vds.go
```diff
@@ -79,6 +85,21 @@ func GetAxis(direction string) (int, error) {
...
+	case "ij":
+		return CoordinateSystemIJ, nil
+	case "ilxl":
+		return CoordinateSystemIlXl, nil
+	case "cdp":
+		return CoordinateSystemCdp, nil
+	default:
+		options := "ij, ilxl, cdp"
+		msg := "coordinate system not recognized: '%s', valid options are: %s"
+		return 0, fmt.Errorf(msg, coordinateSystem, options)
```
ErlendHaa, line 99: As you pointed out in Eirik's pr, returning -1 is a better choice here. GetAxis arguably does the wrong thing. Maybe we can get a tiny fixup commit for that one too ? 😇

achaikou: Ah, I didn't even notice :D

internal/vds/vds.h
achaikou: Unified, but I feel like dividing it into commits (for internal renaming) is more pain then profit.
Not sure if we would like to change public one. But that's a different discussion.

ErlendHaa (approved):

PR #54, opened 2022-10-04
Fetch outer bounds

achaikou:


internal/vds/vds.cpp
```diff
@@ -551,7 +552,12 @@ struct vdsbuffer fetch_fence(
         const float x = *(coordinates++);
         const float y = *(coordinates++);
 
-        const auto coordinate = transform_coordinate(x, y);
+        auto coordinate = transform_coordinate(x, y);
+
+        if (interpolation_method == NEAREST) {
```
ErlendHaa, line 589: This is one of the few cases where I would appreciate a inline comment explaining why it's necessary to do this rounding

internal/vds/vds_test.go
```diff
@@ -425,6 +425,112 @@ func TestFence(t *testing.T) {
...
 }
 
+func TestFenceNearestInterpolationSnap(t *testing.T) {
+	testcases := []struct {
+		coordinate_system int
+		coordinates       [][]float32
+		expected          []float32
+	}{
+		{
+			coordinate_system: CoordinateSystemAnnotation,
```
ErlendHaa, line 490: There should be a case for `x.4999`, together with the `x.5`. As we expect the first to round to `x` and the second to `x + 1`. This boundary is more interesting to test than the `x.25` and `x.75` cases

internal/vds/vds.cpp
```diff
@@ -554,6 +569,23 @@ struct vdsbuffer fetch_fence(
 
         auto coordinate = transform_coordinate(x, y);
 
+        auto validate_boundary = [&] (const int voxel) {
```
ErlendHaa, line 572: The compiler will optimise this, but maybe still move the lambda outside the loop?

internal/vds/vds.cpp
```diff
@@ -554,6 +569,23 @@ struct vdsbuffer fetch_fence(
...
+            const auto min = -0.5;
+            const auto max = layout->GetDimensionNumSamples(voxel_todim(voxel))
+                            - 0.5;
+            if(coordinate[voxel] <= min || coordinate[voxel] >= max) {
```
ErlendHaa, line 576: -0.5 should round "up"  to 0, hence it's in the valid range:
```
if(coordinate[voxel] < min || coordinate[voxel] >= max) {
```

ErlendHaa: Which raises another question. How does the rounding work on negative values?

internal/vds/vds_test.go
```diff
@@ -425,6 +425,61 @@ func TestFence(t *testing.T) {
...
+			coordinates:       [][]float32{{3, 10}, {6, 11.25}},
+			interpolation:     "nearest",
+			error:             "Coordinate (6.000000,11.250000) is out of boundaries in dimension 0.",
+		},
+		{
+			coordinate_system: CoordinateSystemAnnotation,
+			coordinates:       [][]float32{{5.5, 11.5}, {3, 10}},
+			interpolation:     "linear",
+			error:             "Coordinate (5.500000,11.500000) is out of boundaries in dimension 1.",
```
ErlendHaa, line 445: Can we not assert on the numbers in the error message like this? It makes it extremely cumbersome to change stuff

internal/vds/vds_test.go
```diff
@@ -425,6 +425,61 @@ func TestFence(t *testing.T) {
...
+	testcases := []struct {
+		coordinate_system int
+		coordinates       [][]float32
+		interpolation     string
+		error             string
+	}{
+		{
+			coordinate_system: CoordinateSystemAnnotation,
+			coordinates:       [][]float32{{3, 10}, {6, 11.25}},
+			interpolation:     "nearest",
```
ErlendHaa, line 438: Does the interpolation method affect the result in any way? Why does it need to be part of the test case?

internal/vds/vds_test.go
```diff
@@ -425,6 +425,61 @@ func TestFence(t *testing.T) {
 	}
 }
 
+func TestFenceBorders(t *testing.T) {
+	testcases := []struct {
```
ErlendHaa, line 429: The benefit of having a name to a test is that you can explain the contexts of the numbers you are using a bit better. 

E.g. "outside lower bound", "outside upper bound"

And maybe you already do, but it's a bit hard to tell: it makes sense to test the "just out-of-bound" cases. E.g. for index in 'i' direction that would be `-0.49` (lower) and `2.5` (upper)

internal/vds/vds.cpp
```diff
@@ -586,6 +586,10 @@ struct vdsbuffer fetch_fence(
         validate_boundary(0);
         validate_boundary(1);
 
+        /* openvds uses rounding down for Nearest interpolation.
+           As it is counterintuitive, we fix it by snapping to nearest index
+           and rounding halfway cases away from zero.
```
ErlendHaa, line 591: Warning: pedantic comment ahead

Is common to prefix each line with a star
```
/*
 *
 *
 */

File: internal/vds/vds_test.go (hunk 6 above)
Comment (ErlendHaa) [reply internal/vds/vds_test.go:440]: Then I would say that the added complexity/noice outweighs the benefit here. The more variables you add to tests, the less obvious it is what you are testing for

File: internal/vds/vds_test.go [hunk 9]
```diff
@@ -427,40 +427,46 @@ func TestFence(t *testing.T) {
...
 			coordinates:       [][]float32{{700, 1200}},
 			interpolation:     "cubic",
-			error:             "Coordinate (700.000000,1200.000000) is out of boundaries in dimension 0.",
+			error:             "is out of boundaries in dimension 0.",
 		},
 		{
+			name:              "coordinate 2 is just-out-of-lower-bound in direction 1",
 			coordinate_system: CoordinateSystemAnnotation,
-			coordinates:       [][]float32{{3, 11}, {5, 10}, {1, 9.5}},
+			coordinates:       [][]float32{{0.0001, 11}, {5.9999, 10}, {0.0001, 9.5}},
```
ErlendHaa, line 460: Is there any point in have multiple coordinates here? Why not keep it simple and clear by having a single coordinate?

achaikou: Again - slight not really important "randomness". For example, is the first wrong coordinate reported, or is the last one, or are all the ones?
(result was btw way more clear with full error message)

ErlendHaa (approved):

PR #56, opened 2022-11-03
Update openvds

achaikou:


Dockerfile
```diff
@@ -12,9 +12,9 @@ RUN apk --no-cache add \
     util-linux-dev
 
 WORKDIR /
-RUN git clone https://community.opengroup.org/erha/open-vds.git
+RUN git clone https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds.git
 WORKDIR /open-vds
-RUN git checkout enable-cache
+RUN git checkout cbcd7b6163768118805dbcd080a5b0e386b82a6a
```
ErlendHaa, line 17: What's this specific commit?

achaikou: Because this is currently a head and they fixed fmt-issue just two commits ago and not yet released that. Don't think we have to wait until they release a "nice" tag.

ErlendHaa (approved):

Issue #64, opened 2022-11-29
Introduce static golang checker to github actions


achaikou:
https://staticcheck.io/docs/running-staticcheck/ci/github-actions/

Checker seems to find stuff that we miss ans looks to be not too-annoying, so we can test it out.

PR #68, opened 2022-12-01
Static checker fixups

ErlendHaa:
This PR addressed issues reported from the newly introduced golang static checker CI workflow (#64). It's purpose is to make this CI job green.


ajaust (approved):
Looks good to me.

Issue #66, opened 2022-11-29
Refactor the C++ code


ErlendHaa:
There are several issues with the current implementation:

 1)  Inconsistent assertions
We do make some assumption on how the input vds is structured, but this is not consistent across functions. Once we have an explicit list of what we expect #63, we should make constant assertions on our assumptions.

2) Manual coordinate transformations
We do a lot of manual coordinate transformations, mostly in the form of switch-cases. I believe a lot of these are a consequence of us not knowing the OpenVDS library currently at the time. With more experience comes room for improvement and maybe we can offload more to OpenVDS.

3) Separation of concern
Currently we have a "do every-thing at once" design in our end endpoint handler (slice / fence / metadata). There is no clear separation between input validation, VDS file validation, coordinate transform and setup / tear down of required state. 

4) Too much duplication
Currently we have a very linear design where each endpoint handler (slice / fence / metadata) does it's own thing. A lot of that is setup and tear down, as OpenVDS offers a lot of boiler plate. We should aim for a more cohesive implementation where some of that setup and tear down is shared.

Perhaps introducing a stateful "VDS" object can help improve on some of these things

ErlendHaa:
Resolve by #105

Issue #63, opened 2022-11-29
Clarify our expectations to the input VDS


ErlendHaa:
VDS files comes in many forms and shapes. However, our server is designed specifically for seismic. With that comes a some assumptions on how the VDS files we accept should look like. E.g w.r.t. to dimensionality and metadata. 

What we are currently expecting from the input VDS:

- N-dimensions = 3
- Dimension order = Depth/Time/Sample , crossline, inline.
- Annotations are defined
- Coordinate transform (origin, inline step, crossline step) is defined

We should evaluate if there are more expectations we should / make. Generally I would say it's a good thing to start out a bit to0 strict and then loosen up as we see the need. The basic assumptions we make should perhaps be documented in the README and possibly the Swagger docs

ajaust:
I think if we want to be strict we could also think about restrictions/expectations regarding the following points:

- Do we have expectations regarding the number of [channels](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/vds/deepdive/deepdive.html#channels-and-data-type), their order and content? Or more general, what is the [anatomy of the VDS](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/vds/deepdive/deepdive.html#example-anatomy-of-a-vds) what we expect?
  - [SEGYImport](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/tools/SEGYImport/README.html) creates three channels by default containing the cube data, SEGY trace headers and a trace mask channel.
  - Would fast slices be allowed?
- [Compression](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/vds/deepdive/deepdive.html#compression-format-and-quality). Do we allow it?
- [Level of detail](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/vds/deepdive/deepdive.html#level-of-detail-lod) (LOD). Do we allow/expect several LODs? 
  - This could affect the decision on expected [margins](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/vds/deepdive/deepdive.html#margin-size). Do we require the margin to be zero if we expect now LODs?
- [Brick size](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/vds/deepdive/deepdive.html#brick-size). Do we require to the use of the default of uncompressed files?
[...]

PR #69, opened 2022-12-01
Refactoring the C++ internals

ajaust:
This pull request refactors the C++ core of the API which interacts with the OpenVDS library and the library's C++ interface.

## What has been changed (short overview)

- Rename structs, enumerations etc. to use Pascal case. This does not apply to structs in the C-style interface as lower case names for structs are common in C.
- Removed `lineno` from metadata requests of slices since the line number is never used when fetching metadata.
- Bumped C++ requirement to C++17
    - `nlohmann::json` already uses C++17 features.
    - Gives access to improved handling of unique pointers (`std::make_unique`).
- Marking functions, parameters and variables with `noexcept` and `const` where suitable.
- Introduce several new classes
    - `SeismicAxisMap`: Pure virtual base class that makes the indices of inline, crossline and sample axis dimensions used internally by OpenVDS available.
    - `PostStackAxisMap`: Specialization of `SeismicAxisMap` for post-stack datasets.
    - `SeismicHandle`: Simple class for opening a VDS and basic access.
    - `PostStackHandle`: Extension of `SeismicHandle` that deals with post-stack datasts stored as VDS. It contains functionality to request slices, fences etc.
    - `AxisMetadata` and `AxisDescriptor` for accessing metadata from the VDS and additionally accessing axes information and conversions.
- Replaced stray occurences of number like 0, 1, and 2, for example, and replaced them by an enumeration or function call. This helps with understanding what this number means in the context of the function call.
- The substring vds has been removed from functions and function parameters that define the API as much as possible
- Minimized the warnings that are issued if the API is compiled with strict (stricter than standard) warnings enabled.

## What is missing

- Tests of the classes, but not sure if we want/need them at the moment for the classes.
- We could add own exceptions, but I am not sure if there would be much to gain at the moment in terms of reuse of exceptions.

## Questions

- Should we put everything into a namespace called `vds` or similar? I did not use `VDS` in the names of classes since it felt like it led to class names restating the connection to VDS while residing in a directory called `vds/`. I am open for suggestions here.
- We have now inconsistent labels capitalization for enumerations. Enumerations in `vds.h` use all caps labels, while other enumerations capitalize only the first letter. What style do we prefer?
- General remarks, suggestions for improvements, general disagreements?
- What do we do if the coordinate reference system (`crs-wkt`) is not set in the metadata? We could reject the VDS file on opening, but I am not sure if that should be a task of the API nor whether we should rather do this in the future, but not now.

## Remarks

- We could check for attribute names and units of the channel to be correct and sound. What are sound and expected combinations should be discussed and potentially implemented in the future. I will open a new issue or add this to #63.

Related to #66


achaikou (commented):
I feel like I chewed on you quite a bit here.... 
Like a cat on a noble knight that came to slay a dragon. 
But in my defense after this long review the cat became a very confused hungry cat, so no surprise that cat confused self with the dragon! :smile: 


This PR has not been fast & easy to review.  


Reviewer expects "Refactoring" PR to do refactoring only. Reviewer does not expect to see anything new there. Reviewer is very surprised to find new functionality added. :scream_cat: 

Reviewer expects clear history. Reviewer is not happy when having to guess where stuff went, why it went there and why stuff reviewed 10 commits ago is just nuked. In current state of things I lose the main thing I expect to assert when reviewing refactoring PR: nothing is lost, nothing is gained. :crying_cat_face: 

Reviewer gets tired of reviewing 37 commits in one PR. :sleepy: I agree that it is a very hard task to make this refactoring comprehensible in just a couple of clearly divided commits, but this many commits that just mess around in my opinion are even way worse than one big commit "here it is, everything refactored in one step". 

Reviewer is unhappy when is confused too much. And you are very unfortunate :smile: Your reviewer is me and I am easily confused :smile:. Which means that all the code must be so simple, that even I could understand what is going on. And here that was not always the case. 

And unhappy reviewer, by the rules of human nature, becomes an evil reviewer :smiling_imp: 


So I think that the best way forward is to try to unconfuse me! To do that we can: 

- Probably would be for the best to consider this behemoth a prototype and create a couple of new PRs of reasonable size (one with new functionality, one with putting stuff into new structure with minimal changes, one with introducing some other classes, one with some random improvements, etc). 

- Clearly have 1 and only 1 commit per identified change. 

- Get together and draw a diagram of classes that we need and their responsibilities. And important functions for these classes. And also requests flow.
- Get together and discuss some style issues and comments where I am clearly the unreasonable one :smile: 

In my books though you deserve a reward for taking up the task! When faced with tasks like that I just tend to pretend I don't work here :smile:

Issue #70, opened 2022-12-12
Docker container is currently broken due to azure-sdk-for-cpp failing to build


ajaust:
The Docker container fails to build with the following errors

```text
#11 21.29 -- stderr output is:
[1433](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1434)
#11 21.29 /open-vds/3rdparty/azure-sdk-for-cpp-12.3.0/sdk/core/azure-core/src/cryptography/md5.cpp: In member function 'virtual void {anonymous}::Md5OpenSSL::OnAppend(const uint8_t*, size_t)':
[1434](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1435)
#11 21.29 /open-vds/3rdparty/azure-sdk-for-cpp-12.3.0/sdk/core/azure-core/src/cryptography/md5.cpp:166:65: error: 'int MD5_Update(MD5_CTX*, const void*, size_t)' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
[1435](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1436)
#11 21.29   166 |   void OnAppend(const uint8_t* data, size_t length) { MD5_Update(m_context.get(), data, length); }
[1436](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1437)
#11 21.29       |                                                       ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[1437](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1438)
#11 21.29 In file included from /open-vds/3rdparty/azure-sdk-for-cpp-12.3.0/sdk/core/azure-core/src/cryptography/md5.cpp:13:
[1438](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1439)
#11 21.29 /usr/include/openssl/md5.h:50:27: note: declared here
[1439](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1440)
#11 21.29    50 | OSSL_DEPRECATEDIN_3_0 int MD5_Update(MD5_CTX *c, const void *data, size_t len);
[1440](https://github.com/equinor/vds-slice/actions/runs/3675967242/jobs/6216036291#step:3:1441)
[...]

PR #72, opened 2022-12-20
Pin Alpine Linux version of Docker container to Alpine Linux 3.16

ajaust:
The Go container used by the Dockerfile has been updated to use a new release of Alpine Linux (3.17). This breaks the installation procedure of OpenVDS within the container as OpenSSL v3 instead of v1 is used within the new release. Thus we pin the Alpine version to 3.16 which uses OpenSSL v1.


ajaust:
This PR is a temporary fis to #70.

ErlendHaa (approved):

PR #75, opened 2022-12-21
Swap axis order in slice metadata

ErlendHaa:
The axis information does not match the way the data is ordered. One can argue that it's the data that should be transposed, not the axis as that seams to be the most natural orientation for e.g. plotting. However, I'm unsure if we should transpose data server side as that imposed additional compute on us.

Hence this commit reorders the metadata rather than the data, to make sure they correspond. This descision might be revisited before we do a proper versioned release of the API.

### DIFF
--- a/api/query_test.go
+++ b/api/query_test.go
@@ -156,13 +156,13 @@ func TestSliceHappyHTTPResponse(t *testing.T) {
 		switch testcase.slice.Direction {
 		case "i":
 			expectedMetadata = &testSliceMetadata{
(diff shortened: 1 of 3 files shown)


ajaust (commented):
Overall, this looks good to me. Is there any issue or other note to remind us of this issue? Otherwise I would see a risk that we forget about these changes.

ErlendHaa:
I'll make an issue about the potential decision to transpose data server-side

ajaust (approved):

PR #79, opened 2022-12-22
Use an official OpenVDS release tag in Dockerfile

ErlendHaa:
We where checked out at a specific commit due to us relying on a fix in OpenVDS' build system that at the time where not part of any official release. Version 3.1.2 includes the fix in question and we can now use the official release again.


ErlendHaa:
Turns out that there are new build errors related to dependancies in OpenVDS 3.1.2 so we can't really upgrade yet.

PR #81, opened 2022-12-23
Sas srt validation and connection-refactoring

### LINKED ISSUE #80: Safeguard against sas-tokens with "incorrect" allowed resources
Any sas-token which includes "container" but not "blob" in its allowed resources field `srt` case OpenVDS to signal abort and take down the server. This include valid sas tokens. The expected behaviour would be to get access denied (as OpenVDS requires "blob" _and_ "container" to be in `srt`) 

Whether the root cause is in OpenVDS itself or in the azure dependency is unknown.

We should safe-guard against this by manually checking the `srt` field before passing the token to OpenVDS

ErlendHaa:
This PR does 2 things.

The main purpose of this PR is to fix #80. But to get there it first refactors how we handle vds connections in go to a more general approach. This refactoring makes the implementation of the fix more elegant. The design of the new connection is also coloured by some planned expansions needed when implementing caching, which are not part of this PR.

### DIFF
--- a/internal/vds/connection.go
+++ b/internal/vds/connection.go
@@ -6,9 +6,45 @@ import (
 	"net/url"
 )
 
-type Connection struct {
-	Url              string
-	ConnectionString string
+type Connection interface {
+	Url()              string
+	ConnectionString() string
+}
+
+type AzureConnection struct {
+	url              string
+	connectionString string
+}
+
+func (c *AzureConnection) Url() string {
+	return c.url
+}
+
+func (c *AzureConnection) ConnectionString() string {
+	return c.connectionString
+}
+
+func NewAzureConnection(url string, connectionString string) *AzureConnection {
+	return &AzureConnection{
+		url:              url,
+		connectionString: connectionString,
+	}
+}
+
+type FileConnection struct {
+	url string
+}
+
+func (f *FileConnection) Url() string {
+	return f.url
+}
+
+func (f *FileConnection) ConnectionString() string {
+	return ""
+}
+
+func NewFileConnection(path string) *FileConnection {
+	return &FileConnection{ url: path }
 }
 
 /*
@@ -42,7 +78,7 @@ func sanitizeSAS(sas string) string {
 	return strings.TrimPrefix(sas, "?")
 }
 
-type ConnectionMaker func(blob, sas string) (*Connection, error)
+type ConnectionMaker func(blob, sas string) (Connection, error)
 
 func MakeAzureConnection(accounts []string) ConnectionMaker {
 	var allowlist []*url.URL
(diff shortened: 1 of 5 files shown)


internal/vds/connection.go
```diff
@@ -124,6 +157,25 @@ func MakeAzureConnection(accounts []string) ConnectionMaker {
...
+			sanitizeSAS(sas),
+		)
+
+		/*
+		 * OpenVDS (v3.1.2) segfaults on sas-tokens where the srt-field (allowed
+		 * resource types) contains 'c' (container) but _not_ 'o' (object/blob).
+		 * Such as tokens are valid in a general sense, but does not provide enough
+		 * access in order to read a VDS.
+		 *
+		 * Once this bug is fixed upstream this manuall check can go way
```
ajaust, line 173:
```suggestion
		 * Once this bug is fixed upstream this manual check can go way
```
- Small typo
- Is there an issue to remind us of removing the manual valiation?

internal/vds/connection.go
```diff
@@ -6,6 +6,39 @@ import (
...
+}
+
+func srtContainsContainer(srt string) bool {
+	return strings.Contains(srt, "c");
+}
+
+/** Validate 'srt' (Allowed Resource Types)
+ *
+ * Validate that the sas token has both 'c' (container) and 'o' (object/blob)
+ * in it's srt field.
```
ajaust, line 20: Maybe slightly rephrase this to contain information that one needs the `c` and the `o` field for accessing the VDS dataset. 

To me it would make sense to mention here what is checked (str contains `c` and `o`) and what that means (one has access to `container` and object. The reason why we check this could come below where we actually call this function.

ajaust (commented):
Looks good to me overall. I really like how the connection information is very nicely represented by the data structures. I only have minor comments regarding documentation and its phrasing. :)

internal/vds/connection.go
ErlendHaa: See if it's better now! If not I'm happy to give it another go 😄

internal/vds/connection.go
ErlendHaa: typo fixed! No and no, but we should do both

internal/vds/connection.go
```diff
@@ -6,6 +6,39 @@ import (
...
 
+func srtContainsObject(srt string) bool {
+	return strings.Contains(srt, "o");
+}
+
+func srtContainsContainer(srt string) bool {
+	return strings.Contains(srt, "c");
+}
+
+/** Validate 'srt' (Allowed Resource Types)
```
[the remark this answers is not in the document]
ErlendHaa: Absolutely, my bad! I partially blame Microsoft tho, as they refer to it as Allowed Resource Types in the Azure Portal 😄

internal/vds/connection.go
ajaust: Looks good.

internal/vds/connection.go
ErlendHaa: Once merged I'll an issue for us with reference to the code

ajaust (approved):

internal/vds/connection.go
ErlendHaa: Reported upstream and added reminder to remove in our issue-tracker #83

PR #85, opened 2023-01-03
Don't check for 'srt' on Service & User Delegation Sas

ErlendHaa:
This commit makes sure that validateSasSrt does not return an error if 'srt' is missing. This behaviour was never intended and returning an error is considered a bug as both Service- and User Delegation SAS doesn't use this parameters at all. As a consequence requests with such SAS-tokens would fail.


achaikou (approved):
My container sas now works and E2E tests do not fail as well. User delegation stuff should also be good.

PR #89, opened 2023-01-09
Openvds github action image cache

achaikou:
1. To see if main assumption of cache from master being accessible in PR holds, we need to merge first, but at least it holds for stuff pushed to other branches.

2. When openvds must be updated, cache with new version name shouldn't exist anywhere and all the jobs would have to build it on their own. So everything should be fine.

3. Need to push to Radix to make sure dockerfile still builds, but think it should be fine.

### DIFF
--- a/.github/workflows/analyzer.yaml
+++ b/.github/workflows/analyzer.yaml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
       - name: Run check
         run: |
           DOCKER_BUILDKIT=1 docker build -f Dockerfile --target static_analyzer .
--- a/.github/workflows/analyzer.yaml
+++ b/.github/workflows/analyzer.yaml
@@ -14,6 +14,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Load openvds image
+        uses: "./.github/actions/load_openvds_image"
+
       - name: Run check
         run: |
-          DOCKER_BUILDKIT=1 docker build -f Dockerfile --target static_analyzer .
+          DOCKER_BUILDKIT=1 docker build \
+            -f Dockerfile \
+            --build-arg OPENVDS_IMAGE=${{ env.OPENVDS_IMAGE_TAG }} \
+            --target static_analyzer \
+            .
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -12,7 +12,7 @@ jobs:
     name: Run go tests
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Run server tests
         run: |
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -14,6 +14,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Load openvds image
+        uses: "./.github/actions/load_openvds_image"
+
       - name: Run server tests
         run: |
-          DOCKER_BUILDKIT=1 docker build -f Dockerfile --target tester .
+          DOCKER_BUILDKIT=1 docker build \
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -1,4 +1,14 @@
+ARG OPENVDS_IMAGE=openvds
+ARG OPENVDS_REPO=https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds.git
+ARG OPENVDS_VERSION=cbcd7b6163768118805dbcd080a5b0e386b82a6a
+
+FROM alpine as utils
+ARG OPENVDS_VERSION
+RUN echo $OPENVDS_VERSION > /openvds_version.txt
```
ErlendHaa, line 7: It's a bit fragile (and cumbersome like you pointed out) to use the openvds commit sha as cache key in GitHub action. What if something else in the Dockerfile changes? then we don't get a fresh build.

I think it would be simpler to hash the Dockerfile in gh action and use that as key. E.g: `sha1sum Dockerfile`. Then the Dockerfile itself doesn't need to "export" anything at all. radixconfig doesn't care anyway. If we also uaw the same hash as image tag and we should be all set I believe?

.github/workflows/e2e_local.yaml
```diff
@@ -0,0 +1,46 @@
+name: E2E tests with local server
```
ErlendHaa, line 1: Is "local" important information? Every time I read "local e2e test" I need around 5 minutes to figure out what local means in this context. If it's not super relevant I would opt for removing it

.github/workflows/e2e_local.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Load openvds image
+        uses: "./.github/actions/load_openvds_image"
+
+      - name: Build and Run server
+        env:
+          STORAGE_ACCOUNTS: https://${{ secrets.STORAGE_ACCOUNT_NAME }}.blob.core.windows.net
+        run: |
+          tag=server
+          DOCKER_BUILDKIT=1 docker build -f Dockerfile --build-arg OPENVDS_IMAGE=${{ env.OPENVDS_IMAGE_TAG }} --target runner -t $tag .
```
ErlendHaa, line 26: Break long line?

.github/workflows/unit.yaml
```diff
@@ -14,6 +14,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Load openvds image
+        uses: "./.github/actions/load_openvds_image"
+
       - name: Run server tests
         run: |
-          DOCKER_BUILDKIT=1 docker build -f Dockerfile --target tester .
+          DOCKER_BUILDKIT=1 docker build -f Dockerfile --build-arg OPENVDS_IMAGE=${{ env.OPENVDS_IMAGE_TAG }} --target tester .
```
ErlendHaa, line 22: Long line

.github/workflows/analyzer.yaml
```diff
@@ -13,6 +13,10 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
+
+      - name: Load openvds image
+        uses: "./.github/actions/load_openvds_image"
+
       - name: Run check
         run: |
-          DOCKER_BUILDKIT=1 docker build -f Dockerfile --target static_analyzer .
+          DOCKER_BUILDKIT=1 docker build -f Dockerfile --build-arg OPENVDS_IMAGE=${{ env.OPENVDS_IMAGE_TAG }} --target static_analyzer .
```
ErlendHaa, line 22: Question. Is it strictly needed to change the image name, or is it just a nicety ?

ErlendHaa, line 22: Maybe break this line?

achaikou: Not sure what you mean by "change image name"? If you mean that in dockerfile it is named openvds and here it is named openvds:something, then I think it is needed as it must distinguish by name whether image is already there or if it needs to be built.

ErlendHaa (approved):

PR #92, opened 2023-01-12
Radix caching

ErlendHaa:


radixconfig.yaml
```diff
@@ -21,24 +21,22 @@ spec:
...
+          variables:
+            VDSSLICE_CACHE_SIZE: 512 # MB
           monitoring: false
           resources:
             requests:
-              memory: "2048Mi"
-              cpu: "500m"
-            limits:
               memory: "4096Mi"
               cpu: "2000m"
```
achaikou, line 30: And I do think 1000m would make more sense here, but we can decide on the proper resources when going to production.

achaikou (approved):

PR #93, opened 2023-01-13
Log and expose server metrics with prometheus

### LINKED ISSUE #84: Log server statistics with prometheus
It would be useful for us to have better insights of how our endpoints are used by collecting metrics from requests. 

We can do so by setting up a prometheus endpoint (/metrics). There are official golang clients for prometheus [1] that handle the posting of metrics to the endpoint. Radix has built-in support for prometheus (and grafana) [2], and will scrap /metrics out-of-the-box. 

Note that we should also set up /metric on it's own non-private port in the radix config [3]

## Prometheus counters
Here's a (un-complete) list of some metrics that might be useful to keep track of. 
- number of requests
- status codes in responses
- request size
- response size
- request duration 
- cache hit
 
 We probably want all these to be per endpoint. I.e. one individual counter pr endpoint (and http verb?)

[1] https://github.com/prometheus/client_golang
[2] https://www.radix.equinor.com/guides/monitoring/#monitoring-your-app
[3] https://www.radix.equinor.com/references/reference-radix-config/#monitoringconfig

ErlendHaa:
This PR introduces a new endpoint /metrics that is used for serving prometheus metrics. A new internal metrics package is added to contain metrics/prometheus related functionality. Metrics are off by default in our command line app and run on a different port than the rest of the server, as this enables us to keep them private. 

Two prometheus histograms are implemented: 'response size in bytes' and 'duration in seconds', both have filterable labels such as the path (e.g. /slice /fence), method (post, get) and status code (200, 500, etc). Duration also has a cache hit label

closes #84


internal/metrics/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+			ctx.Next()
+			return
+		}
+		
+		start := time.Now()
+		ctx.Next()
+
+		go func() {
+			method := ctx.Request.Method
+			status := http.StatusText(ctx.Writer.Status())
```
achaikou, line 82: Do we want to present status as string or as number? Now it prints "OK", "Not found" etc. Do we maybe want 200/400/404/500 instead?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+const (
+	ms = 0.001
+	s  = 1000 * ms
```
achaikou, line 17: Do I understand correctly that this thing works because s =1 and you give data in seconds? If so, maybe this should be explicitly defined as 1?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+		request_durations: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "vdsslice_durations_histogram_seconds",
+			Help:    "VDSslice latency distributions.",
+			Buckets: []float64{100*ms, 500*ms, 1*s, 2*s, 5*s, 20*s, 1*m, 2*m},
+		}, []string{"method", "path", "status", "cachehit"}),
+
+		response_sizes: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "vdsslice_response_sizes_histogram_bytes",
+			Help:    "VDSslice response size distributions.",
+			Buckets: []float64{100*kb, 1*mb, 5*mb, 10*mb, 20*mb, 50*mb, 100*mb, 200*mb},
```
achaikou, line 56: This thing prints pretty weird stuff like
`status="Internal Server Error",le="1.048576e+06`
This was for wrong inline. Result is 233B.

I think that means that
- we need to add a bucket beneath 1KB.
- 1 unit is a byte, which is good I think, but is there any way that we can format floats without e+06?
- we have to see if there is any way to format the labels, because I would be nice to see "1.1MB" instead of 6 zeroes.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+		}, []string{"method", "path", "status"}),
+	}
+
+	registry.MustRegister(metrics.request_durations)
+	registry.MustRegister(metrics.response_sizes)
+
+	return metrics;
+}
```
achaikou, line 64: Do you know what happens with old data?
Saw somewhere it is removed after 15 days, but it is for Prometheus servers.
Have you maybe seen something like that for our case? Are metrics as long lived as Radix replica?

achaikou (commented):
Nice!! Would be even more nice to send it to test server first before merge. Or after merge...
Think seeing it in grafana can give me a better overview of whats going on.

I am wondering how metrics would look like.
For example, would it be easy to see how many requests are GET - regardless of their size and duration, because I think that is the point of interest. For duration/size I don't care if it is GET or POST request.

Maybe instead of the two metrics we better have more. And duration/size ones then do not need to include Method, for example. And maybe cache hit would have to be its own thing in addition to being included in duration.

internal/metrics/metrics.go
ErlendHaa: Yes I believe so. But logs are cleared on crashes and restarts. But it doesn't matter. A prometheus metric is only a snapshot in time. I.e. "How many request has been processed now". It's graphana that makes the history. It reads the "snapshots" every X seconds and use the difference to provide trends over time. So hopefully it shouldn't matter if logs are reset.

internal/metrics/metrics.go
```diff
@@ -37,6 +61,21 @@ func NewGinMiddleware(metrics *metrics) gin.HandlerFunc {
 			ctx.Next()
 			return
 		}
+		
+		start := time.Now()
+		ctx.Next()
+
+		go func() {
```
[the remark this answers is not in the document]
achaikou: Badly worded/placed from my side.

My main concern was that I couldn't find in the examples/docs for "Observe" what it means. Aka whether when you call "observe" you start an infinite waiting observer that registers a new value once it figures out there is one. 

Now I finally was able to find the line "Observe adds a single observation to the histogram." and I am mostly at peace :smile:

radixconfig.yaml
```diff
@@ -33,8 +40,11 @@ spec:
             maxReplicas: 3
         - environment: test
           variables:
+            VDSSLICE_PORT: 8080
             VDSSLICE_CACHE_SIZE: 256 # MB
-          monitoring: false
+            VDSSLICE_METRICS: true
+            VDSSLICE_METRICS_PORT: 8081
```
achaikou, line 46: Fix replicas too? It needs minimal number of replicas to be 1, not 0.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		request_durations: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "vdsslice_durations_histogram_seconds",
+			Help:    "VDSslice latency distributions.",
+			Buckets: []float64{100*ms, 500*ms, 1*s, 2*s, 5*s, 20*s, 1*m, 2*m},
+		}, []string{"method", "path", "status", "cachehit"}),
+
+		response_sizes: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "vdsslice_response_sizes_histogram_bytes",
+			Help:    "VDSslice response size distributions.",
+			Buckets: []float64{100*kb, 1*mb, 5*mb, 10*mb, 20*mb, 50*mb, 100*mb, 200*mb},
```
achaikou, line 49: Any idea why grafana shows us this
image
when we have way bigger buckets. It shouldn't be able to distinguish between .
Are we wrong in grafana, or here?

I think something is with grafana as when you remove "quantile" you get different results in kilobytes.
But would be nice to confirm.

achaikou (commented):
Go and force-push this thing. I'll see tomorrow the final commits and approve.
Consider if you want to have go tidy somewhere here or not. "No" is also accepted.

---
Though I still want a bar gauge which will show us something like

> response size 100KB - 1MB bucket - average response time is 100ms
> response size 100MB - 200 MB bucket - average response time is 5s

but I can't find how to do that (aka how to join metrics on time and then make average value in one bucket a value in another).

Even if it can't be done in prometeus/grafana, we can add this metric later.

achaikou (approved):

PR #97, opened 2023-01-31
Cache tests

achaikou:
Grew up a bit beyond caching as I discovered new small things to do when trying to figure out what I am doing wrong.

It would be nice to one day discuss and agree on the following things for when we move to production server. For now I chose what I chose, but it is up for discussion.

1. Do we assume our storage accounts are simple storage account, Data Lake Gen 2 or both. Do we set in stone which one we use in e2e tests, or would it be configurable.

2. What would be our test structure to get the best from performance tests

- we now finally got slices with remotely solid time where Erlend said "no, that's not fast enough". Which means there might be something to include into the suite. Currently all the numbers are "this are the numbers we got, because we don't really know what we want"

- we see now that big resources play important role in test results. So we might want to be able to keep running tests on replicas with real resources and do not have dedicated replica with less resources

- see if we would want to exclude requests to "testdata" paths from grafana plots. That would allow us to continue running performance tests on actual production environments when needed without polluting metrics

- how to deal with caching and do we assume it is turned on the instance. As we don't know when instance restart, we can't be sure that request that we make doesn't hit cache, where we actually want to know the time without the cache. We might try to bypass it by restarting dedicated test replica or disabling cache there at all. Or have two dedicated environments - with and without cache. I don't see any other ways that might work

- consider dropping "max" thresholds. We can't really influence it, unless from spinning more replicas. Until we decide that "no request ever should take more than N time", they make little sense. But maybe we will find these boundaries at least for some files?

3. Figure out what is secret. I didn't notice that some variables print in the github logs. Not sure if github won't change anything there in the future without notifying us.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -21,6 +21,9 @@ jobs:
       - name: Build and Run server
         env:
           VDSSLICE_STORAGE_ACCOUNTS: https://${{ secrets.STORAGE_ACCOUNT_NAME }}.blob.core.windows.net
+          # Keep cache size big enough so that cache is never full
+          # thus tests can rely on the key to be accepted on the first try
+          VDSSLICE_CACHE_SIZE: 50
         run: |
           tag=server
           DOCKER_BUILDKIT=1 docker build \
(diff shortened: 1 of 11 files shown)


ErlendHaa (commented):
Other than picking on the python tests a bit, this looks good to me. There is a fair amount of trust from my part when it comes to the JS tests tho 😄

ErlendHaa (approved):

PR #99, opened 2023-02-09
Cleanup of C++ core (part 1 of refactoring)

ajaust:
Overall this PR does some basic cleanup which includes renaming and moving of code into different files. The PR changes the following things:

- Enumeration `axis` is renamed to `api_axis_name`. This frees up the name `Axis` such that it can be used in a follow-up PR. It is also closer to what this enumeration represents: The name of an axis in our API. 
- Moved the `BoundingBox` class into its own header and implementation file. The `vds.cpp` is quite large and follow-up PRs are supposed to introduce additional classes which should be in the own headers in implementations files as well.
- Removes a parameter from functions that obtain metadata for slices as the parameter was never used.
- Rename `vdsbuffer` to `response` as the name better reflects the content of the struct. 

This and follow-up pull requests replace #69.

Next PR is #100.

### DIFF
--- a/internal/vds/vds.cpp
+++ b/internal/vds/vds.cpp
@@ -18,6 +18,8 @@
 #include <OpenVDS/KnownMetadata.h>
 #include <OpenVDS/IJKCoordinateTransformer.h>
 
+#include "boundingbox.h"
+
 using namespace std;
 
 void vdsbuffer_delete(struct vdsbuffer* buf) {
@@ -342,64 +344,6 @@ void set_voxels(
         voxelmax[i] = vmax[i];
     }
 }
-
-
-class BoundingBox {
-public:
-    explicit BoundingBox(
-        const OpenVDS::VolumeDataLayout *layout
-    ) : layout(layout)
-    {
-        transformer = OpenVDS::IJKCoordinateTransformer(layout);
-    }
-
-    std::vector< std::pair<int, int> >       index()      noexcept (true);
-    std::vector< std::pair<int, int> >       annotation() noexcept (true);
-    std::vector< std::pair<double, double> > world()      noexcept (true);
-private:
-    OpenVDS::IJKCoordinateTransformer transformer;
-    const OpenVDS::VolumeDataLayout*  layout;
-};
-
-
-std::vector< std::pair<int, int> > BoundingBox::index() noexcept (true) {
-    auto ils = layout->GetDimensionNumSamples(2) - 1;
-    auto xls = layout->GetDimensionNumSamples(1) - 1;
-
-    return { {0, 0}, {ils, 0}, {ils, xls}, {0, xls} };
-}
-
-std::vector< std::pair<double, double> > BoundingBox::world() noexcept (true) {
-    std::vector< std::pair<double, double> > world_points;
-
-    auto points = this->index();
-    std::for_each(points.begin(), points.end(),
-        [&](const std::pair<int, int>& point) {
-            auto p = this->transformer.IJKIndexToWorld(
-                { point.first, point.second, 0 }
-            );
-            world_points.emplace_back(p[0], p[1]);
-        }
-    );
-
-    return world_points;
-};
-
-std::vector< std::pair<int, int> > BoundingBox::annotation() noexcept (true) {
-    auto points = this->index();
-    std::transform(points.begin(), points.end(), points.begin(),
-        [this](std::pair<int, int>& point) {
-            auto anno = this->transformer.IJKIndexToAnnotation({
-                point.first,
-                point.second,
-                0
-            });
-            return std::pair<int, int>{anno[0], anno[1]};
-        }
-    );
-
-    return points;
-};
 
 nlohmann::json json_axis(
     int voxel_dim,
--- a/internal/vds/vds.cpp
+++ b/internal/vds/vds.cpp
@@ -29,7 +29,7 @@ void vdsbuffer_delete(struct vdsbuffer* buf) {
     *buf = vdsbuffer {};
 }
 
-int axis_todim(axis ax) {
+int axis_todim(axis_name ax) {
     switch (ax) {
         case I:
         case INLINE:
@@ -48,7 +48,7 @@ int axis_todim(axis ax) {
     }
 }
 
-coordinate_system axis_tosystem(axis ax) {
+coordinate_system axis_tosystem(axis_name ax) {
     switch (ax) {
         case I:
         case J:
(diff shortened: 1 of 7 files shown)


internal/vds/vds.h
```diff
@@ -65,13 +64,13 @@ struct vdsbuffer fence(
...
 
-struct vdsbuffer fence_metadata(
+struct response fence_metadata(
     const char* vds,
     const char* credentials,
     size_t npoints
 );
 
-void vdsbuffer_delete(struct vdsbuffer*);
+void response_delete(struct response*);
```
achaikou, line 73: would it make more sense to rename this thing into delete_response?

achaikou (commented):
Nice! I have just two questions which do not necessary should result in changes.
I think we should be able to merge this on Monday!

internal/vds/vds.h
```diff
@@ -37,26 +37,25 @@ enum interpolation_method {
     TRIANGULAR
 };
 
-struct vdsbuffer metadata(
+struct response metadata(
```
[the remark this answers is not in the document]
ajaust: I would say that (to me) camel case is more common in C++, but in general it depends on the project. I decided to stick with the lower case/snake case naming for the following reasons:
- We already use lower case/snake case. Sticking with this creates less noise in the PR and it is consistent with what we have in the implementation file as well. 
  
  In the C++ STL I would also expect lower case names rather than upper case. So there are good examples for any capitalization rule.
- The interface between Go and OpenVDS it technically supposed to be C which might have other conventions than C++. At least I have the impression names to be lower case most of the time in C.

internal/vds/vds.cpp
```diff
@@ -29,7 +29,7 @@ void vdsbuffer_delete(struct vdsbuffer* buf) {
     *buf = vdsbuffer {};
 }
 
-int axis_todim(axis ax) {
+int axis_todim(api_axis_name ax) {
```
ErlendHaa, line 34: Not a fan of the name `api_axis_name` tbh. The `api` part doesn't add anything for me. Maybe if we prefix everything in the C API with `api` it would make sense [1]. But as a lone prefix it's just another meaningless term. After all it doesn't specify what api it refers to. I suggest shortening it too `axis_name`

[1] But then perhaps `api_`is not the best prefix

achaikou: I am more fine with it.
I think the point was to clearly distinguish between things in API and things in openvds.
If we do not have openvds prefix, we need to have an api prefix.
So it is either api, or vdsslice (I've already been through all 5 stages of grief and accepted the name).

I think it should make it either to not mix with axis names from openvds.

internal/vds/vds.h
ErlendHaa: I do not agree 😄 Keep in mind this is a C API. It's a very common pattern in C to have some struct / type and a set of function's to operate on the type (since no classes and whatnot). And when you API have multiple types, you get better grouping of operations (functions) by prefixing the type, not the operation (`type_operation`).

Now we don't have a large C API, not a public one. But my point is that `type_operation` is less supprising, or more common if you will, than `operation_type`. I can live with both though.

achaikou: Now you lie in the commit message! You do more than renaming vds buffer to response!
So maybe either make it a separate commit, or listen and agree with Erlend and leave it as is.

internal/vds/vds.cpp
ajaust: I think that it is useful to have some kind of indication that this naming of the axis is from our side (prefix `api`, `vdsslice`, `openvds_adapter`, `request`...). As we have (at least) two notions of axes: the axes defined within (Open)VDS and in our requests/API. There are quite some cases where the axis do not have the same name, but there has to be a mapping between the axis specified in a request and axes in the VDS dataset that is currently used. 

In the later pull requests, I introduce an axis class that has members to return names/indices for the API/vds-slice (referred to as `api`) naming convention and the OpenVDS (referred to as `vds`) convention. I see that `api` is ambiguous, but it was a tempting term as it is short. 

Would a more expressive name be helpful? Maybe `request_axis_name`?

internal/vds/vds.h
ajaust: I guess I am not completely awake yet. I think I would drop this change for now. If it bothers us a lot in the future, we can still change it then.

internal/vds/vds.cpp
ErlendHaa: We either introduce a consistent naming scheme for referencing our API, or we don't. My point is, this introduces a new term for referencing our C API, but only for a single type. And in the follow-up PR `api` is used once more, but not in the C API that time, then it's on `Axis`.  So it just another noisy word that thrown around. It adds nothing to my understanding, but bloats all code it touches. 

And API is poor name too. (naming is hard, so no judgement). whenever I leave this code for a bit and come back I always have to think: api, which api? oh _our_ api, not openvds's api.

And keeping in mind that in the follow-up PR's we aim to 1) abstract aways OpenVDS into 2 classes and 2) namespace those classes. Which will make the distinction between us and openvds much more obvious. Either you are inside the classes that abstracts openvds, or you are on the outside where openvds "doesn't" exist

ErlendHaa: To be brutally honest I would say that if you still think we need to disambiguate the axis like this after the introduction of `openvds::DataHandle` and `openvds::Metadata`, then those classes have ultimately failed at their goal of abstraction.

ajaust: This is a very strict perspective, but I agree with it. After removing the `vdsbuffer` there are also no (other) argument to keep the distinction by adding prefixes structs/functions. 😄

I will push a fixup that will lead to an out-of-date commit message. The commit message has to be fixed during squashing.

achaikou (commented):
Looks good to me, so wait for Erlend's approval and rebase and merge

ErlendHaa (approved):
Looks good to me!

internal/vds/vds.cpp
```diff
@@ -22,16 +22,16 @@
...
     if (!buf)
         return;
 
     delete[] buf->data;
     delete[] buf->err;
-    *buf = vdsbuffer {};
+    *buf = response {};
 }
 
-int axis_todim(api_axis_name ax) {
```
ErlendHaa: This doesn't belong in this commit

ajaust: My bad. I tried to fix the commit message with a rewording commit. It seems that I messed something up while squashing the rewording. :(

ErlendHaa: Looks good now, merge when green!

PR #107, opened 2023-02-23
Always use R32 as requested data type

### LINKED ISSUE #16: Figure out return data type
Do we always return user data as f4 little-endian?

Do we have any guarantee
1. that returned values from vds are floats (not ints or whatever else was possible in segy)
2. that is always f4, not f8 or some unknown float (which might be possible as openvds tool claims to lose no data)
3. what about endianess and do we care?
4. what does it all has to do with `enum class VolumeDataFormat` ?

Right now
- we return format for slice endpoint (metadata "format")
- return nothing for fence (it is supposed to be float, but is it always the same float or not? What if VolumeDataTraces are not float?).

achaikou:
closes #16


internal/vds/metadatahandle.cpp
```diff
@@ -81,10 +81,15 @@ std::string MetadataHandle::crs() const noexcept (true) {
 }
 
 std::string MetadataHandle::format() const noexcept (false) {
-    auto format = this->m_layout->GetChannelFormat(0);
+    /*
+     * We always request data in OpenVDS::VolumeDataFormat::Format_R32 format
+     * as this seems to be intended way when working with openvds.
+     * Thus users will always get data returned as f4.
+     *
+     * We also assume that server code is run on a little-endian machine.
```
ErlendHaa: I would love a link to the issue on gitlab where this is explained in more detail

ErlendHaa (commented):
LGTM

ErlendHaa (approved):

PR #110, opened 2023-02-28
Make lines in well-defined file perpendicular

achaikou:
Apparently xlines and ilines are always expected to be perpendicular. This requirement wasn't known for previous version of well-defined file.

Note: well-defined file must be updated on cloud right before merging to make e2e pass.


ErlendHaa (approved):
🙏

Issue #122, opened 2023-04-19
Expose the "ImportTimeStamp" field from VolumeDataLayout through /metadata


ErlendHaa:

AdilKhurshid:
close by #125

PR #125, opened 2023-04-20
Added display name in metadata response

AdilKhurshid:
Issue: https://github.com/equinor/vds-slice/issues/122

### DIFF
--- a/internal/vds/metadatahandle.hpp
+++ b/internal/vds/metadatahandle.hpp
@@ -19,6 +19,7 @@ public:
 
     BoundingBox bounding_box() const noexcept (true);
     std::string crs()          const noexcept (true);
+    std::string input_filename() const noexcept (true);
 
     Axis const& get_axis(Direction const direction) const noexcept (false);
 
--- a/internal/vds/vds.cpp
+++ b/internal/vds/vds.cpp
@@ -313,6 +313,7 @@ struct response metadata(
     nlohmann::json meta;
 
     meta["crs"] = metadata.crs();
+    meta["inputFileName"] = metadata.input_filename();
 
     auto bbox = metadata.bounding_box();
     meta["boundingBox"]["ij"]   = bbox.index();
--- a/internal/vds/vds.go
+++ b/internal/vds/vds.go
@@ -104,18 +104,18 @@ type FenceMetadata struct {
 
 func GetAxis(direction string) (int, error) {
 	switch direction {
-		case "i":         return AxisI,         nil
-		case "j":         return AxisJ,         nil
-		case "k":         return AxisK,         nil
-		case "inline":    return AxisInline,    nil
-		case "crossline": return AxisCrossline, nil
-		case "depth":     return AxisDepth,     nil
-		case "time":      return AxisTime,      nil
-		case "sample":    return AxisSample,    nil
-		default:
-			options := "i, j, k, inline, crossline or depth/time/sample"
-			msg := "invalid direction '%s', valid options are: %s"
-			return -1, fmt.Errorf(msg, direction, options)
+	case "i":         return AxisI,         nil
+	case "j":         return AxisJ,         nil
+	case "k":         return AxisK,         nil
+	case "inline":    return AxisInline,    nil
+	case "crossline": return AxisCrossline, nil
+	case "depth":     return AxisDepth,     nil
+	case "time":      return AxisTime,      nil
+	case "sample":    return AxisSample,    nil
+	default:
+		options := "i, j, k, inline, crossline or depth/time/sample"
+		msg := "invalid direction '%s', valid options are: %s"
+		return -1, fmt.Errorf(msg, direction, options)
 	}
 }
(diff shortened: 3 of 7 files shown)


internal/vds/metadatahandle.hpp
```diff
@@ -19,6 +19,7 @@ class MetadataHandle {
 
     BoundingBox bounding_box() const noexcept (true);
     std::string crs()          const noexcept (true);
+    std::string get_displayname() const noexcept (true);
```
ErlendHaa, line 22: I know the above methods (`crs` and `bounding_box`) and `get_axis` below follow different conventions on using a `verb` so you fall in-between to chairs here. I'd opt for dropping get her tho, as the new method is conceptually closer to the ones above.

ErlendHaa (commented):
Nice! Tests are failing. Have a look the build logs in GitHub Actions or run them locally (`go test -race ./...` and update the failing test :)

internal/vds/vds.cpp
```diff
@@ -308,7 +308,8 @@ struct response metadata(
     nlohmann::json meta;
 
     meta["crs"] = metadata.crs();
-
+    meta["displayName"] = metadata.get_displayname();
```
ErlendHaa, line 311: This is 100% my fault, but after looking at the OpenVDS docs I think we want [InputFileName](https://osdu.pages.opengroup.org/platform/domain-data-mgmt-services/seismic/open-vds/cppdoc/class/classOpenVDS_1_1KnownMetadata.html#_CPPv4N7OpenVDS13KnownMetadata30ImportInformationInputFileNameEv) instead.

AdilKhurshid: Sure! I replaced the display name with the input file name and function

internal/vds/metadatahandle.hpp
AdilKhurshid: I changed the function name to `input_filename`. Let me know if it follows the convention we are using or just suggest me the appropriate name.

ErlendHaa (commented):
Looks good, but you also need to update the Metadata struct in `vds.go` which serves as documentation in swagger

internal/vds/vds.go
```diff
@@ -109,10 +110,10 @@ func GetAxis(direction string) (int, error) {
...
 		case "time":      return AxisTime,      nil
 		case "sample":    return AxisSample,    nil
-		default:
-			options := "i, j, k, inline, crossline or depth/time/sample"
-			msg := "invalid direction '%s', valid options are: %s"
-			return -1, fmt.Errorf(msg, direction, options)
+	default:
+		options := "i, j, k, inline, crossline or depth/time/sample"
+		msg := "invalid direction '%s', valid options are: %s"
+		return -1, fmt.Errorf(msg, direction, options)
```
ErlendHaa, line 116: This indentation is technically correct. But the change is incomplete. The cases should also be un-indented:
```
switch ...
case ...
case ...
default:
```

It's greatly encourage to fix stuff like this when you find it. Even when it's unrelated to the PR. From experience, that is the only way it's ever getting fixed. Nobody is gonna bother to create issues and stand-alone PRs just with minor fixed like this. But fixing them as part of some other work, like you do here is fast and simple. BUT in order to not distract from what's really going on in commit, these types of fixes should be granted there own small commits like: "Fix incorrect indentation in switch-case" or "Fix whitespace errors". That reduces noice in the diffs of the other commits. 

So complete the indentation and extract that (together with the other indentation changes) in a separate commit. And continue fixing stuff like this when you find it 🎉

ErlendHaa (commented):
Looks good. Fix up and squash the history and do a force-push to get a clean history

internal/vds/vds.go
AdilKhurshid: Thanks for letting me. I didn't add this intentionally. It's the IDE who reformatted the file indentation and I mistakenly push that in the PR. It's good that you mentioned it, I will take care of that next time and add additional commits whenever needed. :)

internal/vds/vds.go
```diff
@@ -80,6 +78,9 @@ type Metadata struct {
 	// Coordinate reference system
 	Crs string `json:"crs" example:"PROJCS[\"ED50 / UTM zone 31N\",..."`
 
+	// The original input file name
+	InputFileName string `json:"inputFileName" example:"EQ21200DZC21B-KPSDM-RAW-FULL-0535-DEPTH.sgy"`
```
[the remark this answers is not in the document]
AdilKhurshid: Done

internal/vds/vds.go
ErlendHaa: Tip: I _always_ use `git add -p`. Then I see exactly what I add to each commit 👍

ErlendHaa (approved):
LGTM, merge away 🎉

AdilKhurshid:
> LGTM, merge away 🎉

Can you merge this PR? I don't have write access to this repository.

PR #130, opened 2023-04-24
Horizon afterfixes

achaikou:


internal/vds/regularsurface.hpp
```diff
@@ -80,7 +80,9 @@ class RegularSurface{
         if (row >= this->nrows()) throw std::runtime_error("Row out of range");
         if (col >= this->ncols()) throw std::runtime_error("Col out of range");
 
-        return this->m_transformation.to_world(col, row);
+        Point point {static_cast<double>(row), static_cast<double>(col)};
```
ErlendHaa, line 95: Maybe overkill, but this static_casting got me thinking:

```
template< typename T >
struct Point {
    T x;
    T y;
}
```

and 
```
template< typename T >
Point< double > operator*(Point< T > p) const noexcept (true) {
```
Which means we can multiply the size_t directly:

```
return this->m_transformation * Point(row, col);
```

ErlendHaa (approved):
LGTM 👍

internal/vds/regularsurface.hpp
achaikou: Looking at static cast had exactly same thought yesterday :handshake: 
The code is nice. What stopped me though is the supposed generality o AffineTransformation.

Does it make sense to accept a point of whatever kind and return double? It suddenly stops being symmetrical and thus becomes a bit off from mathematical concept.
And as we don't have very large rows/columns we won't anyway lose precision on converting.
So decided not to ponder a lot and use static_cast.

What do you think? Should we make it nicer for us instead?

PR #139, opened 2023-05-16
Use float as initial value in rms

achaikou:
Same issue as in 7e22dd04865406aef76518898c89114991fef8c4 Our tests were based on integers, thus error was not caught.


ErlendHaa (approved):
Ideally this should be 2 commit, but otherwise looks good

achaikou:
I think on the opposite it is expected to fix the bug and the tests that didn't catch the bug in one commit.

PR #144, opened 2023-06-02
Calculate attributes concurrently

ErlendHaa:


internal/vds/vds.go
```diff
@@ -503,24 +512,56 @@ func (v VDSHandle) GetAttributes(
...
-		C.float(above),
-		C.float(below),
-		C.float(stepsize),
-		unsafe.Pointer(&buffer[0]),
-	)
 
-	if err := v.Error(cerr); err != nil {
-		return nil, err
+	maxConcurrency := 32
+	windowsPerRoutine := int(math.Ceil(float64(hsize) / float64(maxConcurrency)))
```
achaikou, line 519: Question: do you think there would be a performance problem anywhere here with jumping logic?
If hsize = 96, we have 32 threads with each window size 3.
If hsize = 97, suddenly we have 25 threads - 24 with window size 4 and one with window size 1.

I am wondering if it would be worse, better or the same to keep number of threads consistent?
Like if hsize = 97, have 32 threads - 31 with window size 3 and 1 with window size 4.

Second approach probably complicates logic, but I can't really tell if it would matter.

internal/vds/vds.go
```diff
@@ -420,6 +421,14 @@ func (v VDSHandle) GetFenceMetadata(coordinates [][]float32) ([]byte, error) {
 	return buf, nil
 }
 
+func min(a, b int) int {
```
achaikou, line 424: (pure go complaint, you are good)

all I have to say: :woman_facepalming: 

(should we hide it as a local function maybe?)

internal/vds/vds.go
```diff
@@ -503,24 +512,56 @@ func (v VDSHandle) GetAttributes(
...
+
+		size := min(windowsPerRoutine, remaining)
+		to := from + size
+
+		go func(from, to int) {
+			var cCtx = C.context_new()
+			defer C.context_free(cCtx)
+
+			cErr := C.attribute(
+				cCtx,
```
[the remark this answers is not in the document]
ErlendHaa: I should of course not use v.Error here 🤦

internal/vds/vds.go
ErlendHaa: I know right! I cannot believe I had to write that method. It's already local to this file.

internal/vds/vds.go
ErlendHaa: From what I've tested it's definitly not worth it. The current worst case is that the last routine only gets 1 window. I.e. it will finish right away an in practise we are utilising one less go routine. On my machine it doesn't seam to do much if I double or half the numbers of goroutines, so I don't think this will matter much

internal/vds/vds.go
```diff
@@ -503,24 +512,56 @@ func (v VDSHandle) GetAttributes(
...
+	remaining := hsize;
+	fmt.Println(hsize)
+	nRoutines := 0;
+	for remaining > 0 {
+		nRoutines++
+
+		size := min(windowsPerRoutine, remaining)
+		to := from + size
+
+		go func(from, to int) {
```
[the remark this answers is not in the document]
ErlendHaa: I think I will sit down and do a full refactoring of GetAttributes later to clean it up a bit

achaikou (approved):

PR #146, opened 2023-06-05
Attribute request refactoring

ErlendHaa:


api/request.go
```diff
@@ -263,7 +263,7 @@ func (h AttributeRequest) Hash() (string, error) {
 func (h AttributeRequest) toString() (string, error) {
 	msg := "{vds: %s, Rotation: %.2f, Origin: [%.2f, %.2f], " +
 		"Increment: [%.2f, %.2f], FillValue: %.2f interpolation: %s, " +
-		"Above: %f, Below: %f, Stepsize: %f, Attributes: %v}"
+		"Above: %2f, Below: %2f, Stepsize: %2f, Attributes: %v}"
```
achaikou, line 236: What's the difference between %2f and %.2f ?
I see only dot in the examples.

api/request.go
```diff
@@ -139,9 +139,9 @@ func (s SliceRequest) toString() (string, error) {
 	return str, nil
 }
 
-// Query for Horizon endpoints
-// @Description Query payload for horizon endpoint /horizon.
-type HorizonRequest struct {
+// Query for Attribute endpoints
+// @Description Query payload for attribute endpoint.
+type AttributeRequest struct {
```
achaikou, line 144: Also note that now user sees horizon endpoint under horizon tag, but is supposed to use AttributeRequest to create the request itself.
I think there is no link that Horizon endpoint relates to AttributeRequest in any way.

So not sure if AttributeRequest shouldn't be HorizonRequest or we should find another way to map them.

api/request.go
ErlendHaa: Should be dot

api/request.go
ErlendHaa: Unsure, but the way the swagger docs are layed out I don't think this will be a source of much confusion

achaikou (approved):

PR #147, opened 2023-06-05
Iterator typedef + docs

ErlendHaa:


internal/vds/attribute.hpp
```diff
@@ -135,6 +135,12 @@ class AttributeMap {
 public:
     AttributeMap(void* dst, std::size_t size) : dst(dst), size(size) {};
 
+    /** Iterator type for computations
+     *
+     * All attributes computes produce a float, but the computation itself
+     * should happen on double's to not loose precision during intermediate
+     * computation steps.
```
achaikou, line 142: All attribute() computes
should happen on double()s to not lo()se

PR #149, opened 2023-06-14
0.5 adjustment and lack of proper interpolation tests

achaikou:


internal/vds/vds.cpp
```diff
@@ -416,32 +380,28 @@ void fetch_horizon(
...
-
-            if (not inrange(iline, ij[0]) or not inrange(xline, ij[1])) {
+            auto ij = transform.WorldToOpenvdsPosition(
+                {cdp.x, cdp.y, OpenvdsVoxelVector::NOVALUE}
+            );
+            auto k  = transform.AnnotationToOpenvdsPosition(
+                {OpenvdsVoxelVector::NOVALUE, OpenvdsVoxelVector::NOVALUE, depth}
+            );
+
+            auto coordinate = OpenvdsVoxelVector(k[0], ij[1], ij[2]);
```
ErlendHaa, line 390: Just curious, have you tested the performance of this?

The old call stack for these 3 was:
```
transform.WorldToIJKPosition
transform.AnnotationToIJKPosition
```

Now it's
```
transform.WorldToOpenvdsPosition
  WorldToIJKPosition
    OpenVdsVoxelVector
      OpenVDS::DoubleVector3 (copy)

transform.AnnotationToOpenvdsPosition
  AnnotationToIJKPosition
    OpenVdsVoxelVector
      OpenVDS::DoubleVector3 (copy)

OpenvdsVoxelVector
  OpenVDS::DoubleVector3 (copy)
```

for every point in the horizon

ErlendHaa: That seams like a totally reasonable size. 

What is that ,  ~15-25% increase in execution time? Then we have to ask ourselves. Do we accept a hit like that in order to remove some ugly, but not magic, numbers from the code?

achaikou: To answer that question I need a handover from you on where to get something similar to actually desired environment and how you yourself use it!

I think below certain threshold we wouldn't care much if something increases in size.
Like if some operation contributes to the final result 1ms, but after a change would contribute 2ms, that is probably OK.

So after some point the absolute value matters more than the relative one. We need to figure out if we would reach that point of not on proper hardware.

If my machine is 4 times slower then final one, the difference for this horizon would be around-ish 0.1 second.
If user pattern doesn't expect user to request 100s horizons per second, that might it theory be acceptable, if it benefits us enough.

internal/vds/vds.cpp
```diff
@@ -252,16 +252,6 @@ void fetch_fence(
...
-        /* OpenVDS' transformers and OpenVDS data request functions have
-         * different definition of where a datapoint is. E.g. A transformer
-         * (To voxel or ijK) will return (0,0,0) for the first sample in
-         * the cube. The request functions on the other hand assumes the
-         * data is located in the center of a voxel. I.e. that the first
-         * sample is at (0.5, 0.5, 0.5). This is a *VERY* sharp edge in the
-         * OpenVDS API and borders on a bug. It means we cannot directly
-         * use the output from the transformers as input to the request
-         * functions.
-         */
```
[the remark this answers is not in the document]
ErlendHaa: You haven't moved it anywhere in this commit ?

internal/vds/metadatahandle.hpp
```diff
@@ -23,6 +26,43 @@ class MetadataHandle {
...
+        OpenvdsVoxelVector ToOpenvdsPosition(
+            const OpenVDS::DoubleVector3& coordinate,
+            enum coordinate_system coordinate_system
+        ) const;
+
+        bool IsOpenvdsPositionOutOfRangeIline (double value) const noexcept(true);
+        bool IsOpenvdsPositionOutOfRangeXline (double value) const noexcept(true);
+        bool IsOpenvdsPositionOutOfRangeSample(double value) const noexcept(true);
+        bool IsOpenvdsPositionOutOfRange   (const OpenvdsVoxelVector& position) const noexcept(true);
+        Axis WhereOpenvdsPositionOutOfRange(const OpenvdsVoxelVector& position) const noexcept(false);
```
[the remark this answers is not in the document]
ErlendHaa: >Wanted to have everything that extends some openvds functionality to have this prefix to make it obvious that we extend something that shouldn't have been extended, this is not our class.

A noble attempt, but you can't make this definition without documenting it. If that naming convention is supposed to add something to all devs it must be known to all devs. 

Personally I don't see the point in trying to distinguish stuff from openvds with names. Everything in openvds lives under a namespace anyway.

My complaints are made, do with it what you want 😄

internal/vds/metadatahandle.hpp
```diff
@@ -23,6 +26,43 @@ class MetadataHandle {
...
+            OpenVDS::IJKCoordinateTransformer(metadata.m_layout), m_metadata(metadata) {};
+
+        OpenvdsVoxelVector IndexToOpenvdsPosition     (const OpenVDS::DoubleVector3& position) const;
+        OpenvdsVoxelVector AnnotationToOpenvdsPosition(const OpenVDS::DoubleVector3& position) const;
+        OpenvdsVoxelVector WorldToOpenvdsPosition     (const OpenVDS::DoubleVector3& position) const;
+
+        OpenvdsVoxelVector ToOpenvdsPosition(
+            const OpenVDS::DoubleVector3& coordinate,
+            enum coordinate_system coordinate_system
+        ) const;
```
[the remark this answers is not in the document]
ErlendHaa: We've added them to all classed that are "sorta library" ish, to indicate usage. Either we use it everywhere or we can remove. But I think we decided we wanted them. If we are not positive openvds can't throw it must be `noexcept (false)`

internal/vds/metadatahandle.hpp
```diff
@@ -23,6 +26,43 @@ class MetadataHandle {
...
+     * bridge this gap.
+     */
+    class CoordinateTransformer : public OpenVDS::IJKCoordinateTransformer{
+    public:
+        CoordinateTransformer(const MetadataHandle& metadata ) :
+            OpenVDS::IJKCoordinateTransformer(metadata.m_layout), m_metadata(metadata) {};
+
+        OpenvdsVoxelVector IndexToOpenvdsPosition     (const OpenVDS::DoubleVector3& position) const;
+        OpenvdsVoxelVector AnnotationToOpenvdsPosition(const OpenVDS::DoubleVector3& position) const;
+        OpenvdsVoxelVector WorldToOpenvdsPosition     (const OpenVDS::DoubleVector3& position) const;
```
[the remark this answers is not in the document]
ErlendHaa: I have 2 😄 

All of these is a wrapper around one specific method on the openvds transformer. By naming them exactly the same it's easy to understand which is which. The methods are anyway on a different class in a different namespace than the originals so I don't see any problems with reusing the same names

Or I would use the fact that you return a `VoxelVector` and just name these `fromIJK`, `fromAnnotation` and `fromWorld`, but I like my first suggestion better 😄

internal/vds/openvdsvoxelvector.hpp
```diff
@@ -0,0 +1,62 @@
...
+     Voxel position 0       Voxel position 1               Voxel position nsamples-1
+          Sample 0               Sample 1                     Sample nsamples-1
+
+ * -> Shows how coordinate is stored in OpenvdsVoxelVector and its relation to actual
+ * sample positions and results from ToVoxelIndex/ToVoxelPosition functions.
+ *
+ * In contrast to openvds data in the vector are always stored in
+ * Inline/Crossline/Sample order.
+ */
+class OpenvdsVoxelVector : public OpenVDS::DoubleVector3{
```
ErlendHaa, line 30: I'm not super chuffed about this class. I guess I fail to see the justification for adding a wrapper around the point class. Probably because for me this entire endouvior is about the mismatch between the output of OpenVDS's transformers and what it expects in its access manager. Hence, what makes sense in my head is to make a wrapper around the open-vds transformer so that we can add the 0.5 offset. But here we have 2 wrappers for one thing and I don't follow why we want that?

internal/vds/metadatahandle.hpp
```diff
@@ -23,7 +27,44 @@ class MetadataHandle {
...
+     * We have a problem that OpenVDS' ToVoxelPosition and data request
+     * functions have different definitions of where a datapoint is. E.g. A
+     * transformer will return (0,0,0) for the first sample in the cube. The
+     * request functions on the other hand assume the data is located in the
+     * center of a voxel. I.e. that the first sample is at (0.5, 0.5, 0.5).
+     *
+     * Thus extending IJKCoordinateTransformer with new methods helps us to
+     * bridge this gap.
+     */
+    class CoordinateTransformer : public OpenVDS::IJKCoordinateTransformer{
```
ErlendHaa, line 43: Why is it a nested class ?

internal/vds/openvdsvoxelvector.hpp
```diff
@@ -0,0 +1,62 @@
...
+     */
+    OpenvdsVoxelVector(
+        const MetadataHandle& metadata,
+        const OpenVDS::DoubleVector3& original,
+        const OpenVDS::DoubleVector3& calculated
+    ) :
+        OpenVDS::DoubleVector3(
+            original[0] == NOVALUE ? NOVALUE : calculated[metadata.iline().dimension()]  + 0.5,
+            original[1] == NOVALUE ? NOVALUE : calculated[metadata.xline().dimension()]  + 0.5,
+            original[2] == NOVALUE ? NOVALUE : calculated[metadata.sample().dimension()] + 0.5
```
ErlendHaa, line 52: Conceptually I still think this belongs in the transform-class

internal/vds/openvdsvoxelvector.hpp
ErlendHaa: > One of these issues is a DoubleVector. It is just a data array, with no meaning behind it. To know what it contains and be able to use it correctly you need to know which function created that double vector. Was that to IJKIndex or was that to VoxelPosition? In theory a separate class prevents you from using any DoubleVector in a place where I actually want only those "Double Vectors which represent coordinates to Request functions".

Sounds like you want a simple typedef maybe?

> I think this is a quite frequent problem: should this datastructure that I use stay raw, or should I add a light wrapper over it for convenience?

True and to answer that we weigh up the pros vs cons. I.e. added complexity vs convenience. So then my question is, what convenience does it add? We can now do
`voxel.InlineValue()` instead of `voxel[inline.dimension()]`.  Which maybe looks a bit nicer, but introduces tighter coupling. Personally I think the complexity over adding a new class and having to understand how it related to openvds out weighs the benefit here. 

But consider this as questions more than request for change

internal/vds/metadatahandle.hpp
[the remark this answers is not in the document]
ErlendHaa: My initial thought was to just override the methods where we wanted to change the behaviour (Because I'm considering the openvds transformer to be wrong). So conceptually something like this:

```
struct Transformer : private OpenVDS::IJKCoordinateTransformer {
    using base_type = OpenVDS::IJKCoordinateTransformer;
     
    // Expose all the methods we don't want to change
    using base_type::base_type; // constructor

    using base_type::IJKPositionToAnnotation;
    using base_type::IJKIndexToAnnotation;
    ....

    // Custom implementation where needed
     DoubleVector3 WorldToVoxelPosition(DoubleVector3 const& world) const noexcept (false) {
         return base_type::WorldToVoxelPosition(world) + 0.5;
    } 
}
```

This keeps the original interface and avoid the problem of coming up with new names. Feel free to disregard if you don't like it

internal/vds/metadatahandle.hpp
```diff
@@ -23,7 +27,50 @@ class MetadataHandle {
...
+            OpenVDS::IJKCoordinateTransformer(metadata.m_layout), m_metadata(metadata) {};
+
+        OpenvdsVoxelVector IndexToOpenvdsPosition(
+            const OpenVDS::DoubleVector3& position
+        ) const noexcept (false);
+        OpenvdsVoxelVector AnnotationToOpenvdsPosition(
+            const OpenVDS::DoubleVector3& position
+        ) const noexcept (false);
+        OpenvdsVoxelVector WorldToOpenvdsPosition(
+            const OpenVDS::DoubleVector3& position
```
ErlendHaa, line 55: I'm regretting this so much, but I think we decided to go with east const

internal/vds/metadatahandle.hpp
ErlendHaa: >IJKCoordinateTransformer needs to know about layout which belongs to MetadataHandler.
MetatadataHandler needs to know about CoordinateTransformer and needs to create it.
I want CoordinateTransformer to know about axes, aka it needs to know about MetadataHandler.

This is very I start to think we have a too complicated, entangled and rigid dependency graph between objects

internal/vds/metadatahandle.hpp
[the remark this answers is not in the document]
ErlendHaa: Yes! I commented with an example on on another conversation. Resolving this and continuing the discussion over there

PR #150, opened 2023-06-15
Return geospacial information in slice metadata

ErlendHaa:
Geospacial information is often integral in order to relate seismic to other data sources. Currently we provide the bounding box of the entire cube through /metadata. However, when fetching a (vertical) slice the user would have to manually do coordinate mapping to get the geospacial position of the slice. This commit adds this information to the slice metadata. For horizontal slices, this is just the bounding box, while for vertical slices this is essentially a linestring.

There are many ways to implement this, and I landed on basing it on the SubVolume bounds. This has 2 benefits. Firstly, it ensures consistency between fetch_slice and fetch_slice_metadata, and secondly, it is future proof if we ever deside to implement sub-slicing.

### DIFF
--- a/internal/vds/vds.cpp
+++ b/internal/vds/vds.cpp
@@ -152,6 +152,78 @@ nlohmann::json json_axis(
     return doc;
 }
 
+nlohmann::json json_slice_geospatial(
+    MetadataHandle const& metadata,
+    Direction const direction,
+    Axis const& axis,
+    int lineno
+) {
+    auto const& transformer = metadata.coordinate_transformer();
+
+    SubVolume bounds(metadata);
+    bounds.set_slice(axis, lineno, direction.coordinate_system());
+
+    auto const lower = transformer.VoxelIndexToIJKIndex({
+        bounds.bounds.lower[0],
+        bounds.bounds.lower[1],
+        bounds.bounds.lower[2]
+    });
+
+    // The upper bound is exclusive, while we need it to be inclusive
+    auto const upper = transformer.VoxelIndexToIJKIndex({
+        bounds.bounds.upper[0] - 1,
+        bounds.bounds.upper[1] - 1,
+        bounds.bounds.upper[2] - 1
+    });
+    
+    /** The slice bounds are given by the lower- and upper-coordinates only:
+     *
+     *
+     *       Depth / Time slice   Inline slice   Crossline slice
+     *       ------------------   -------------  ---------------
+     *
+     *         3         upper       upper
+     *         +-----------+           +
+     *         |           |           |
+     *         |           |           |        lower       upper
+     *         |           |           |          +-----------+
+     *         |           |           |
+     *         |           |           |
+     *   J     +-----------+           +
+     *   ^   lower         1         lower
+     *   |
+     *   +--> I
+     *
+     * For inline- and crossline-slices the horizontal bounding box is given by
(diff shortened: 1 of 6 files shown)


internal/vds/vds_test.go
```diff
@@ -389,6 +389,79 @@ func TestSliceMetadataAxisOrdering(t *testing.T) {
...
+			expected: [][]float64{{8.0, 4.0}, {6.0, 7.0}},
+		},
+		{
+			name: "Crossline",
+			direction: AxisCrossline,
+			lineno: 10,
+			expected: [][]float64{{2.0, 0.0}, {14.0, 8.0}},
+		},
+		{
+			name: "J",
```
achaikou, line 418: That method is understandably a bit imprecise. 
```
    auto wp1 = ijkCoordinateTransformer.WorldToVoxelIndex({12, 10, -16});
    std::cout << "Voxel index of world 12 10 -16: "<<wp1[0] << " " << wp1[1] << " " << wp1[2] << "\n";
    auto wp2 = ijkCoordinateTransformer.VoxelIndexToWorld(wp1);
    std::cout << "World of voxel index before: "<<wp2[0] << " " << wp2[1] << " " << wp2[2] << "\n";
```

gives me
```
Voxel index of world 12 10 -16: 3 1 2
World of voxel index before: 12 11 -16
```

Should we mention it nearby that "not all testcases, if you want to implement more, would be this nice"? Like J=1 already breaks the stuff

ErlendHaa: What is imprecise? Why is it understandably so? And what does `j=1` break?   If I change the test to `J = 1` I get `[]float64{0, 3}, []float64{12, 11}`, which is correct

internal/vds/vds.cpp
```diff
@@ -203,30 +203,61 @@ void fetch_slice_metadata(
...
+     *         |           |           |
+     *         |           |           |
+     *   J     +-----------+           +
+     *   ^   lower         1         lower
+     *   |
+     *   +--> I
+     *
+     * For inline- and crossline-slices the horizontal bounding box is given by
+     * a linestring from (lower.I, lower.J) to (upper.I, upper.J). However, for
+     * time- and depth-slices the we need to construct 4 corners. The first
```
achaikou, line 240: (the?) we need

internal/vds/vds.cpp
```diff
@@ -203,30 +203,61 @@ void fetch_slice_metadata(
...
+     * Corner 1 is given by (upper.I, lower.J) and corner 3 is 
+     * (lower.I, upper.J).
+     */
     const std::array< OpenVDS::DoubleVector3, 4 > corners {{
-        transformer.VoxelIndexToWorld({ lower[0], lower[1], lower[2] }),
-        transformer.VoxelIndexToWorld({ lower[0], lower[1], upper[2] }),
-        transformer.VoxelIndexToWorld({ upper[0], upper[1], upper[2] }),
-        transformer.VoxelIndexToWorld({ upper[0], upper[1], lower[2] }),
+        transformer.IJKIndexToWorld({ lower[0], lower[1], 0 }),
+        transformer.IJKIndexToWorld({ upper[0], upper[1], 0 }),
```
achaikou, line 247: Happy that at least one test failed!

achaikou (approved):

PR #151, opened 2023-06-16
Consistently use Go module "require" for testing

ErlendHaa:
When we introduced the module to the repository for the first time, we didn't go back and update old tests. Hence we have a good mix of manually comparisons and the use of require. 

To keep life simple for us, I've also removed all uses of the module assert.

### DIFF
--- a/cmd/query/main_test.go
+++ b/cmd/query/main_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 
 	"github.com/gin-gonic/gin"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -619,12 +618,12 @@ func TestLogHasNoSas(t *testing.T) {
 
 		requireStatus(t, testcase, w)
 
-		assert.NotContainsf(t, buffer.String(), "sas",
+		require.NotContainsf(t, buffer.String(), "sas",
 			"Test '%v'. Log should not contain SAS (sas)", testcase.base().name)
 		// just in case also check for presence of parts of the token
-		assert.NotContainsf(t, buffer.String(), "se=",
+		require.NotContainsf(t, buffer.String(), "se=",
 			"Test '%v'. Log should not contain SAS (se=)", testcase.base().name)
-		assert.NotContainsf(t, buffer.String(), "se%3D",
+		require.NotContainsf(t, buffer.String(), "se%3D",
 			"Test '%v'. Log should not contain SAS (encoded se=)", testcase.base().name)
 	}
 }
@@ -639,7 +638,7 @@ func testErrorHTTPResponse(t *testing.T, testcases []endpointTest) {
 		err := json.Unmarshal(w.Body.Bytes(), testErrorInfo)
 		require.NoError(t, err, "Test '%v'. Couldn't unmarshal data.", testcase.base().name)
 
-		assert.Containsf(t, testErrorInfo.Error, testcase.base().expectedError,
+		require.Containsf(t, testErrorInfo.Error, testcase.base().expectedError,
 			"Test '%v'. Error string does not contain expected message.", testcase.base().name)
 	}
 }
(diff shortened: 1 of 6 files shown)


cmd/query/main_test.go
```diff
@@ -619,12 +618,12 @@ func TestLogHasNoSas(t *testing.T) {
 
 		requireStatus(t, testcase, w)
```
achaikou, line 620: Maybe also update utils_main_test.go? It has some "if err != nil"

internal/vds/vds_test.go
```diff
@@ -83,15 +82,11 @@ func TestMetadata(t *testing.T) {
...
+	require.NoErrorf(t, err, "Failed to retrive metadata, err %v", err)
 
 	var meta Metadata
 	err = json.Unmarshal(buf, &meta)
-	if err != nil {
-		t.Fatalf("Failed to unmarshall response, err: %v", err)
-	}
+	require.NoErrorf(t, err, "Failed to unmarshall response, err: %v", err)
 
 	if !reflect.DeepEqual(meta, expected) {
```
achaikou, line 91: DeepEqual probably can go too?

internal/vds/vds_test.go
```diff
@@ -635,16 +498,10 @@ func TestInvalidFence(t *testing.T) {
...
-		if err.Error() != expected {
-			msg := "Unexpected error message, expected \"%s\", was \"%s\""
-			t.Errorf(msg, expected, err.Error())
-		}
-	}
+	require.ErrorContainsf(t, err,
+		"invalid coordinate [1 1 0] at position 1, expected [x y] pair",
+		"Incorrect error msg",
+	)
 }
```
achaikou, line 506: To "TestFenceInterpolation"
exchange a lot with require.NotEqual ?

internal/vds/vds_test.go
```diff
@@ -689,19 +546,16 @@ func TestFenceInterpolationDefaultIsNearest(t *testing.T) {
...
```
achaikou, line 547: The one above is fine, if you don't want to change it, but not sure if you wanted to but didn't go through the whole file.

internal/vds/vds_test.go
```diff
@@ -767,9 +614,9 @@ func TestSurfaceUnalignedWithSeismic(t *testing.T) {
 		interpolationMethod,
 	)
 	require.Len(t, buf, len(targetAttributes), "Wrong number of attributes")
-	require.NoError(t, err, "Failed to fetch horizon")
+	require.NoErrorf(t, err, "Failed to fetch horizon")
 	result, err := toFloat32(buf[0])
-	require.NoError(t, err, "Failed to covert to float32 buffer")
+	require.NoErrorf(t, err, "Failed to covert to float32 buffer")
 	require.Equalf(t, expected, *result, "Horizon not as expected")
 }
```
achaikou, line 622: Update "TestSurfaceWindowVerticalBounds" too?

ErlendHaa: I don't want to use `require.Equal(t, err, testcase.Err)` here if that's what you mean. Then I have to hardcode the exact error message into each test case

internal/vds/vds_test.go
```diff
@@ -573,7 +572,7 @@ func TestOnly3DSupported(t *testing.T) {
 		handle.Close()
 	}
 
-	require.ErrorContainsf(t, err, "3 dimensions, got 4", "Incorrect error msg")
```
achaikou: why did you remove "got 4"?

achaikou (commented):
Get ready for merge.
And see if you want to fix some of the remaining results when searching for t.Error and t.Fatal

internal/vds/vds_test.go
ErlendHaa: Because my vim macro was incorrect 😄 I'll fix it on squash

achaikou (approved):

PR #152, opened 2023-06-16
Horizon performance tests

achaikou:


tests/performance/helpers/horizon-helpers.js
```diff
@@ -0,0 +1,81 @@
...
+  let horizon = new Array(rows).fill(rowArray);
+  console.log(`Requesting horizon of size ${columns * rows}`);
+  return sendHorizonRequest(horizon, surface, above, below, stepsize, attributes);
+}
+
+/**
+ * Sends horizon request of size rows/columns where every point is at random depth.
+ * Every point is expected to have data without need for interpolation in any dimension.
+ */
+export function sendRandomHorizonRequest(
```
ErlendHaa, line 69: Randomness in a performance test, sure that is a good idea?

achaikou: Absolutely.
I don't care what depth is requested as long as I can assure that
1) we are not likely to hit cache
2) we do not request flat-ish surface and end up downloading too little data

Point 2 might be good to have if that is a real user pattern, but as I am in the process of losing faith in real tests ability to catch degrading performance, there seem to be no point in making it more complicated then necessary.

ErlendHaa (approved):

PR #166, opened 2023-07-07
Split c and c++

achaikou:
Having a detour in writing c++ tests discovered that I want another detour aka for example being able to find data returned by the fetch_horizon.
Hence split.

Also did another thing that's been getting on my nerves for far too long - aka `#include "vds.h"` in simple classes.

Also hope that renaming will help me in no longer opening wrong file (vds.go/vds.cpp) as we can't separate them another way without using cmake to build library separately.


ErlendHaa (commented):
I like the new capi header and implementation files, although I don't buy your argument about cdata.h why cdata must be it's own file. That's just how includes work in c/cpp, but oh well. Not too fan of the name `operations` either. Is `process` better? As in "process request"

ErlendHaa (approved):
LGTM

PR #167, opened 2023-07-12
Rename vds package to core

achaikou:
Changes are brought to you by "Rename Symbol" vscode command.


ErlendHaa (commented):
LGTM, but I would prefer if you separate formatting in its own commit. It becomes much harder to review a commit when when formatting changes are merged together with the actual changes you are trying to do.

internal/core/vds.go
```diff
@@ -108,6 +108,16 @@ type FenceMetadata struct {
 	Shape []int `json:"shape" swaggertype:"array,integer" example:"10,50"`
 } // @name FenceMetadata
 
+// @Description Attribute metadata
+type AttributeMetadata struct {
```
ErlendHaa: Reminds me, maybe we could make a general struct with format and shape. I think these are used for all 3 endpoints? Mostly to centralise docs.

```
type Array struct {
    Format string ...
    Shape []int
}

type AttributeMetadata struct { 
    Array
}
```

docs/attribute.md
```diff
@@ -38,14 +38,14 @@ sumneg      | Sum of negative samples
 
 ## Response
 On success (200) the multipart/mixed response consists of n parts. The first
-part is json document with metadata about the attributes.Each of the next n - 1
-parts contains one attribute. The ordering of the attributes from the request
-is preserved.
+part is json document with metadata about the attributes. Each of the next n - 1
```
ErlendHaa, line 41: ... is _a_ json ...

ErlendHaa (approved):
LGTM

Now you need to update our unofficial client in the demo repo: https://github.com/equinor/oneseismic-demos/blob/c84427d7736f0301936bb50411ab7810250607bc/oneseismic.py#L89

PR #175, opened 2023-07-28
Split RegularSurface.hpp

achaikou:


internal/core/regularsurface.hpp
```diff
@@ -62,7 +62,7 @@ class RegularSurface{
     {}
 
     /* Grid position (row, col) -> world coordinates */
-    Point coordinate(
+    Point toCDP(
```
ErlendHaa, line 65: `toCDP` is follows the golang naming scheme. A more c++-ish name would be `to_cdp`

internal/core/attribute.cpp
```diff
@@ -236,7 +236,7 @@ void calc_attributes(
     auto fill = horizon.fillvalue();
 
     for (std::size_t i = from; i < to; ++i) {
-        auto depth = surface.at(i);
```
ErlendHaa, line 239: For reference: the name `at` is taken from the standard library containers (vector, array). They define `at` as "access specified element with bounds checking".

achaikou: Yes, you said so. And because our structure is not a simple array, I always was confused about what it would return. Like does .at(i) return me a value of the cell or a coordinate of the cell?
I still consider RegularSurface to be a somewhat weird structure, but I think for now "value" fits better than "at".

ErlendHaa (approved):
`to_cdp` sound like a conversion to me. But it's not in out case, it's a lookup / indexing. Do with that what you want 😄

Issue #173, opened 2023-07-21
Fix attributes error handling


achaikou:
We have an unfortunate behavior in our "dead" code.

When calculating attributes we make sure that every single thread returned [without an error](https://github.com/equinor/vds-slice/blob/0dc640c3086ce2b53ab050a3e3e8d610facf2470/internal/core/core.go#L592). That is a logical thing to do in theory.
In practice, however, we could never truly trigger this behavior as there is nothing in our control that could cause an exception down the line.
That's why this piece of code had to stay untested.

This piece of code is however wrong.

1. For demo purposes spoil our code with throwing an exception in some case when calculating attributes.
2. Make an attributes call
3. Observe the server crashing.

What we want instead is a correct handling of the request and propagation of (probably just any if there are many) exception to the user.

PR #176, opened 2023-07-31
Wait for all goroutines to complete before returning

ErlendHaa:
This commit fixes an bug in VDSHandle.GetAttributes that causes server crashes when c.attributes return a nonzero error code.

What happens is that GetAttributes fires of multiple goroutines, each calling C.attribute on parts of the data. Each goroutine reports errors back to the main thread on a error channel. The main thread is then blocked until it gets as many responses as it sent of goroutines. Unless there is an error, then it returns immediately. And that's the bug. The early return means that go free's all of GetAttributes' local variables. Local variables that are still referenced and used in C by the not-yet-completed goroutines. And the server goes boom...

The fix is simply to gather up any errors and delay the return until all goroutines have completed.

Currently, there is no way for our C.attribute function to return a nonzero error code. Atleast, not when called from
VDSHandle.GetAttributes as that does all nessesary input checks before calling the c function. That's why this error has passes under the radar and also why this patch doesn't include any tests.

### DIFF
--- a/internal/core/core.go
+++ b/internal/core/core.go
@@ -592,11 +592,17 @@ func (v VDSHandle) GetAttributes(
 		remaining -= size
 	}
 
+	// Wait for all gorutines to finish and collect any errors
+	var computeErrors []error
 	for i := 0; i < nRoutines; i++ {
 		err := <- errs
 		if err != nil {
-			return nil, err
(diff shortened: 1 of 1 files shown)


ErlendHaa:
Closes #173

achaikou (approved):

PR #180, opened 2023-08-01
Cache entry size

### LINKED ISSUE #168: Incorrect return value for CacheEntry.Size()
This is buggy: 

https://github.com/equinor/vds-slice/blob/35df1fef29f984a0d2d035475878006d5fe46956/internal/cache/cache.go#L27C63-L27C63

Specifically `len(c.data)`

Interestingly enough the test case we have for this does not pick it up. So that is buggy too

### LINKED ISSUE #169: Cachesize is limited to 4GB
Using `uint32` here https://github.com/equinor/vds-slice/blob/35df1fef29f984a0d2d035475878006d5fe46956/internal/cache/cache.go#L108 mean that the max size for our cache is 4GB, which is not a lot. Should change the type to `uint64` or possibly `int64` as that is what ristretto is using

AdilKhurshid:
closes #169
closes #168

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -54,7 +54,7 @@ func (c *RistrettoCache) Get(key string) (val CacheEntry, hit bool) {
 	return val, hit;
 }
 
-func NewRistrettoCache(cacheSize uint32) *RistrettoCache {
+func NewRistrettoCache(cacheSize uint64) *RistrettoCache {
 	/**  Maxcost and NumCounters
 	 *
 	 * This Ristretto cache is configured with a max size in bytes (the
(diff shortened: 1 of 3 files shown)


internal/cache/cache.go
```diff
@@ -24,7 +24,11 @@ func (c *CacheEntry) Metadata() []byte {
 }
 
 func (c *CacheEntry) Size() int {
-	return len(c.data) + len(c.metadata) + int(unsafe.Sizeof(*c))
+	var dataLength int
+	for _, val := range c.data {
+        dataLength += len(val)
```
ErlendHaa, line 29: Weird indent, did `go fmt` do this? 😄

internal/cache/cache_test.go
```diff
@@ -23,13 +23,15 @@ func TestRistrettoCacheMaxSize(t *testing.T) {
...
 	 * 48                   + 512             + 464                 = 1024
 	 */
-	entry := NewCacheEntry(
-		make([][]byte, 512),
-		make([]byte, 464),
-	)
+	data := make([][]byte, 4)
+	for i := range data {
+        data[i] = make([]byte, 128)
+    }
```
ErlendHaa, line 29: Indent

ErlendHaa (approved):
Looks good to me, but the commit `Change cache size data type from uint32 -> unit64` is a fixup commit to the other commit and should be squashed before merging

PR #183, opened 2023-08-01
Regularsurface as parameter

achaikou:
RegularSurface is displayed in the docs as `core.RegularSurface`, but I don't see how to fix that.

### DIFF
--- a/api/request.go
+++ b/api/request.go
@@ -185,31 +185,8 @@ func (s SliceRequest) toString() (string, error) {
 type AttributeRequest struct {
 	RequestedResource
 
-	// Horizon / height-map
-	Horizon [][]float32 `json:"horizon" binding:"required"`
-
-	// Rotation of the X-axis (East), counter-clockwise, in degrees
-	Rotation *float32 `json:"rotation" binding:"required"`
-
-	// X-coordinate of the origin
-	Xori *float32 `json:"xori" binding:"required"`
-
-	// Y-coordinate of the origin
-	Yori *float32 `json:"yori" binding:"required"`
-
-	// X-increment - The physical distance between columns in horizon
-	Xinc float32 `json:"xinc" binding:"required"`
-
-	// Y-increment - The physical distance between rows in horizon
-	Yinc float32 `json:"yinc" binding:"required"`
-
-	// Any sample in the input horizon with value == fillValue will be ignored
-	// and the fillValue will be used in the amplitude map.
-	// I.e. for any [i, j] where horizon[i][j] == fillValue then
-	// output[i][j] == fillValue.
-	// Additionally, the fillValue is used for any point in the horizon that
-	// falls outside the bounds of the seismic volume.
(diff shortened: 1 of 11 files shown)


internal/core/core.go
```diff
@@ -51,6 +51,35 @@ type Axis struct {
...
+	Yinc float32 `json:"yinc" binding:"required"`
+
+	// Any sample in the input horizon with value == fillValue will be ignored
+	// and the fillValue will be used in the amplitude map.
+	// I.e. for any [i, j] where horizon[i][j] == fillValue then
+	// output[i][j] == fillValue.
+	// Additionally, the fillValue is used for any point in the horizon that
+	// falls outside the bounds of the seismic volume.
+	FillValue *float32 `json:"fillValue" binding:"required"`
+}
```
ErlendHaa, line 81: make this `} // @name RegularSurface` to set the display name in the docs

tests/performance/helpers/horizon-helpers.js
```diff
@@ -23,6 +30,7 @@ export function sendHorizonRequest(
   let payload = Object.assign({}, properties, surface);
   sendRequest("horizon", payload);
 }
+//yaata
```
ErlendHaa, line 33: yaata! baata! kayomba!

api/request.go
```diff
@@ -185,31 +185,8 @@ func (s SliceRequest) toString() (string, error) {
...
-
-	// Any sample in the input horizon with value == fillValue will be ignored
-	// and the fillValue will be used in the amplitude map.
-	// I.e. for any [i, j] where horizon[i][j] == fillValue then
-	// output[i][j] == fillValue.
-	// Additionally, the fillValue is used for any point in the horizon that
-	// falls outside the bounds of the seismic volume.
-	FillValue *float32 `json:"fillValue" binding:"required"`
+	// Surface along which data must be retrieved
+	Surface core.RegularSurface `json:"surface" binding:"required"`
```
ErlendHaa, line 189: This breaks the API. Are we sure we want that? We would have to inform all users and update all demo scripts.

internal/core/core.go
```diff
@@ -51,6 +51,35 @@ type Axis struct {
 	Unit string `json:"unit" example:"ms"`
 } // @name Axis
 
+// @Description Geometrical plane with depth/time datapoints
+type RegularSurface struct {
+	// Horizon / height-map
+	Horizon [][]float32 `json:"horizon" binding:"required"`
```
ErlendHaa, line 57: _if_ we break the API, maybe we want to use that opertunity to rename this field too? As it's inside a structure now, maybe `values` or `data` is better? xtgeo uses `values`

api/request.go
ErlendHaa: Ignore the above comment, I remember why this is _needed_ for future API additions. However, I'm missing some motivation for this change in the commit and also more explicitness on the fact that we are breaking the API here, not just combining some structs internally

ErlendHaa (approved):
LGTM

PR #190, opened 2023-08-10
Generalize horizon

achaikou:
Part 2 of "between surfaces". This PR changes underlying horizon/attributes handling to make it work with data of different sizes.

Testing attributes from c++ is quite inconvenient as one has to create too many entities.
We'll test it from go next when it is exposed it.
It means there still could be some bugs in current implementation, but probably not huge as all current horizon tests pass, so difference in heights would be the only new feature.

Compared this branch to master.
Today it seems like new solution is around-ish 2% slower. (Or these could be fluctuations between runs, because next time master was 2% slower...).
Seems acceptable to me if other option is maintaining 2 different implementations.


internal/core/attribute.cpp
```diff
@@ -239,6 +239,12 @@ void calc_attributes(
     AttributeMap::AttributeComputeParam params;
 
     for (std::size_t i = from; i < to; ++i) {
+        auto above = primary.value(i) - top.value(i);
+        auto below = bottom.value(i) - primary.value(i);
+
+        src_window.move(above, below);
+        dst_window.move(above, below);
```
ErlendHaa, line 238: Now that we have a top and bottom surface it seams like the current design for windows (using above below) is wrong, and actually complicates things.  Instead what makes more sense now is looking at the window as a range, from some depth to some depth:

```
const float from = top.value(i);
const float to   = bottom.value(i);

auto window = Window(from, to);
```

This (together with striding and margins, like before) gives us everything we need.

This also removes the need to pass in 3 surfaces

internal/core/cppapi_data.cpp
```diff
@@ -246,14 +272,12 @@ void horizon(
     RegularSurface const& surface,
     float above,
     float below,
+    std::size_t* data_offset,
```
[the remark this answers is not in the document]
ErlendHaa: Scratch that last comment, as we are not passing the horizon buffer straight to the attributes. Instead we are passing the re-sampled temporaries. Which are of `std::vector< double >`, meaning that `std::distance` would be constant and we can already get rid of the size parameter without a performance hit

internal/core/core.go
```diff
@@ -298,29 +298,35 @@ func (r *CRegularSurface) Close() error {
 	return nil
 }
 
-func (surface *RegularSurface) ToCRegularSurface() (CRegularSurface, error) {
-	var cCtx = C.context_new()
-	defer C.context_free(cCtx)
-
+func (surface *RegularSurface) ToCRegularSurface(apply func(float32, int, int) C.float) (CRegularSurface, error) {
```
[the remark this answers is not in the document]
achaikou: (Forgot to unexport CRegularSurface. Would do in separate commit)

internal/core/attribute.cpp
```diff
@@ -30,33 +30,42 @@ Horizon::Window Horizon::at(std::size_t i) const noexcept (false) {
...
 
+void AttributeMap::AttributeComputeParam::update(
+    InputIt begin,
+    InputIt end,
+    std::size_t size,
+    std::size_t reference_index
+) noexcept (true) {
+    this->begin = begin;
+    this->end = end;
+    this->size = size;
```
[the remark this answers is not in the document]
ErlendHaa: > but is there a benefit of getting rid of size parameter and using std::distance instead?
Less variables floating around

Your call tho

internal/core/attribute.cpp
```diff
@@ -30,33 +30,42 @@ Horizon::Window Horizon::at(std::size_t i) const noexcept (false) {
     };
 }
 
+void AttributeMap::AttributeComputeParam::update(
+    InputIt begin,
+    InputIt end,
```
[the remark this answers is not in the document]
ErlendHaa: But that logic every function should be `T some_func(some_func_params const&)`. It's a poor design because it becomes less clear what arguments you expect (you have an indirection in order to figure it out). I see how it's _nessesary_ / nice in order to unify attribute::calculate signatures, but no need to over do it.

internal/core/capi.h
```diff
@@ -123,19 +123,19 @@ int fence_metadata(
...
 
 int horizon(
     Context* ctx,
     DataHandle* handle,
-    RegularSurface* surface,
-    float above,
-    float below,
+    RegularSurface* primary,
+    RegularSurface* top,
+    RegularSurface* bottom,
```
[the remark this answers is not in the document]
ErlendHaa: >In general, primary is still the same "data" which we had in previous horizons. Your reference points.

Why isn't it called reference then?

>And I like passing the same 3 surfaces around in all 3 calls :) Seems clear to me.

You just said that one of them changes definition based the context in which the original request was formed. A context which is lost at this point. Opposite of clear 😄 

> Remember that when we are interpolating we need to pass down the reference point - the one we anchor data with. And for "nearest" calculations we also need a reference point.


Sure, but that doesn't need to be the middle/reference point. It might as well be the top point.

> Not exactly sure we want to do it, but it wasn't part of the task anyway.

If you set out on a task to unify two concept, then I would argue it _is_ a part of the task to make abstraction that makes sense for the new unified codebase. Not just jamming new stuff into old concepts. (VerticalWindow was terrible before, but in this context it's even worse)

internal/core/attribute.cpp
ErlendHaa: > You still need a reference point. Remember that we need to recalculate sampling rate from SOME point. So you always want you 3 surfaces.

You need _a_ reference point, not necessarily the middle point. We used the point from the input surface because that was the only point we had before. Now you can choose either for the 3 point (top, ref, bottom).

internal/core/core.go
```diff
@@ -570,38 +584,33 @@ func (v VDSHandle) GetAttributes(
 		cAttributes[i] = C.enum_attribute(targetAttributes[i])
 	}
 
-	cPrimarySurface, err := primarySurface.ToCRegularSurface(nil)
+	cPrimarySurfaceData, err := primarySurface.toCdata(0)
```
ErlendHaa, line 587: What is the benefit of having both `toCdata` and `toCRegularSurface` over just

`func newCRegularSurface(surface RegularSurface, shift float) (CRegularSurface, error)`

(If it's because it gets big, you can split out the C array stuff in its own helper function that's used internally)

To make caller code as nice as possible I would preferably have an overload like this too:

```
func newCRegularSurface(surface RegularSurface) (CRegularSurface, error) {
    return newCRegularSurface(surface, 0)
}
```

internal/core/attribute.hpp
```diff
@@ -150,6 +150,10 @@ class AttributeMap {
     struct AttributeComputeParam{
         InputIt begin;
         InputIt end;
+        /* size could be exchanged with std::distance, but only as long as
+         * InputIt come from std::vector, so complexity is constant. Passed here
```
ErlendHaa, line 145: > but only as long as InputIt come from std::vector

The requirement is not that it's a vector< T >::iterator, It's that the iterator itself is a random access iterator. Which is true for many types of iterators (std::array for example). Or even our own strides one can be made a random access iterator. Essentially the iterator needs to implement `operator += ` to qualify

internal/core/attribute.cpp
achaikou: No!! :scream_cat: 

We used it as a reference point because that was the only reference point that made sense! We discussed it!!

Why would we change the behavior to something wrong/more complicated/inconsistent with how horizon works now? The idea of this PR was to make minimal changes to allow the Attributes-between-surfaces case, but still keeping horizon intact.

Top and bottom surface and align horizontally, but not as per stepsize, not supplied, not the one from the file.

If your file stepsize is 4, your above is 7 and you new samplesize is 3, there is a big difference which point would be a reference one. For example when interpolating you would get two totally different results if you
- set reference point to main surface. Then you interpolate data to depth, depth - 3, depth - 6.
- set reference point to top surface (aka main + above). Then you interpolate data to top, top - 3, top - 6. You never get data from your supplied surface point as it is top - 7.

Obviously we want to take reference points from primary surface supplied by user. So if they supply a point that has data in the file we return as samplevalue that exact data, and not some estimate.

ErlendHaa: You are right, I forgot about unalignement of input above / below. This is why I don't like to unify things to early, I'm to stupid to keep all the possible conditions and outcomes in mind at the same time

internal/core/attribute.cpp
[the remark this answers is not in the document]
achaikou: I've tried removing it and I do not like it. I think if you provide parameters as an object class, it irritates me less than when parameters were provided on construction and were visible in signature.

Now it is hidden and can be just considered calculation, like storing something in a variable because it is nicer than calculating the same value in every loop iteration.

So leaving as is.

internal/core/capi.h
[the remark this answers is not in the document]
ErlendHaa: Trust me, I'm not all all attached to VerticalWindow 😂 My point is that now that we have surfaces for above/below I _think_ we can solve the task that VerticalWindow solves, using only these surfaces. But we leave it for the future.

internal/core/core.go
ErlendHaa: Gad damn golang, why you no have overloads?

internal/core/attribute.hpp
```diff
@@ -145,7 +138,21 @@ class AttributeMap {
      */
     using InputIt = std::vector< double >::iterator;
 
-    virtual float compute(InputIt begin, InputIt end) noexcept (false) = 0;
+    struct AttributeComputeParam{
```
ErlendHaa, line 141: ..Params ?

ErlendHaa (approved):
LGTM! Rebase away

PR #194, opened 2023-08-17
Attributes between surfaces

achaikou:
All the changes needed to make attributes between surfaces available to the users and some minor convenient side-changes.

After this is merged we will update the demo scripts, push to the Radix test branch and let PO check solution on real files.
At the same time we will inform current users about breaking changes, which we make as we still are not in production. Only then we will push to the main branch.


internal/core/core_test.go
```diff
@@ -1470,6 +1471,101 @@ func TestAttribute(t *testing.T) {
...
+				0.000001,
+				"[%s]\nExpected: %v\nActual:   %v",
+				targetAttributes[i],
+				expected[targetAttributes[i]],
+				*result,
+			)
+		}
+	}
+}
+
```
achaikou, line 1568: TODO: missing a test exclusively consisting of fillvalues to emulate splitting where one thread gets fillvalues only.

internal/core/core_test.go
```diff
@@ -2204,6 +2204,59 @@ func TestAttributesInconsistentLength(t *testing.T) {
 	require.ErrorContains(t, err, errmsg, err)
 }
 
+func TestAttributesAllFill(t *testing.T) {
+	const above = float32(0)
+	const below = float32(0)
+	const stepsize = float32(4)
+	targetAttributes := []string{"samplevalue", "min"}
```
AdilKhurshid, line 2211: Is there any specific reason that we are only testing `samplevalue` and `min` attributes? If not, would it be nicer to check all the attributes here?

achaikou: There is no real point in testing more.
Test makes sure that if you supply only fillvalue values, all works nicely and you get fillvalues in return.

As we do not have harsh testing requirements, we try to implement only tests that give best value - balance between actually having functionality tested and not overblowing test suite too much.

This is a whitebox test, we know the system, so we can use our knowledge that all attributes are actually implemented similarly and used attributes should have no meaning for the test outcome. The number of attributes shouldn't technically matter as well, but due to having to fill the same result for every attribute we better test on >1. Minimal necessary number is 2.

Hence I just chose two example attributes here and they happened to be samplevalue and min.

AdilKhurshid:
As I do not know the complete context of the entire task which this PR is solving but I guess you have already discussed these changes with, Erland So it looks good to me as well. I'm Approving this PR 🎉 😊

AdilKhurshid (approved):

PR #202, opened 2023-09-12
Add requirements step to Readme e2e guide

yngve793:
Describe how to setup the python environment for the e2e tests.


achaikou (approved):
Go and merge! We usually just click Merge so it creates a Merge commit and keeps history the way it was during review.

PR #211, opened 2023-10-03
Endpoint duplication

achaikou:
I've been irritated for a while that our endpoints are copied whenever we need a new one.
This attempts to extract away currently duplicated logic.

### DIFF
--- a/api/endpoint.go
+++ b/api/endpoint.go
@@ -91,7 +91,7 @@ func (e *Endpoint) slice(ctx *gin.Context, request SliceRequest) {
 		return
 	}
 
-	cacheKey, err := request.Hash()
+	cacheKey, err := request.hash()
 	if abortOnError(ctx, err) {
 		return
 	}
@@ -140,7 +140,7 @@ func (e *Endpoint) fence(ctx *gin.Context, request FenceRequest) {
 		return
 	}
 
-	cacheKey, err := request.Hash()
+	cacheKey, err := request.hash()
 	if abortOnError(ctx, err) {
 		return
 	}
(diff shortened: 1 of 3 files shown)


yngve793 (commented):
The last two commits contains restructuring of Go code. I will leave the review of this part to Adil as he is more experienced in Go.

api/endpoint.go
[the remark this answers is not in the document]
achaikou: :woman_facepalming: 
Of course you are right, thank you.
That's what you get on discovering right before push that function doesn't actually need one parameter...

api/endpoint.go
```diff
@@ -122,110 +122,64 @@ func (e *Endpoint) makeDataRequest(
...
-	}
-	defer handle.Close()
-
+func (request SliceRequest) execute(
+	handle core.VDSHandle,
+) (data [][]byte, metadata []byte, err error) {
 	axis, err := core.GetAxis(strings.ToLower(request.Direction))
-	if abortOnError(ctx, err) {
+	if err != nil {
 		return
```
AdilKhurshid, line 130: I am not completely sure but just to confirm, Will it return a proper error message in case of an error?

Is it not supposed to be like this 🤔

`return nil, nil, err `

achaikou: I originally had it like
`return nil, nil, err `

But there are so many returns here, that it just becomes annoying.
So I decided to use go feature [Named return values](https://go.dev/tour/basics/7)

Wouldn't use it in simple functions, as in my mind they harm readability, but here the number of returns warrants it. 
It also has some additional consequences. Like if there error is in data function, it will return `nil, metadata, error` instead of nil, nil, error. But it should be ok as far as I can see.
Also if err is not nil it will return whatever err is, so we should be safe.

Haven't use this feature before, so we can see if it was a mistake to use it long-term?

AdilKhurshid (approved):

yngve793 (approved):
No further comments.

PR #214, opened 2023-10-13
Add position of minimum value attribute

### LINKED ISSUE #198: Add attributes for time/depth position of max, min and maxabs
Introduce the following attributes:
```
max_at
min_at
maxabs_at
```
They are supposed to return respective depth/time value of the max, min and maxabs attributes.

In case there are several depth/times with the same value, we could return any, but it must be consistent on every call.

AdilKhurshid:
Added a new attribute 'min_at' that will return the position of the minimum value in sample axes for both along the surface and between the surfaces.

Note:

This is not the final PR, but the implementation for only the minimum value position. If we agree with these changes, then I will add the rest of the attributes, which are similar to this.

closes #198

### DIFF
--- a/internal/core/attribute.cpp
+++ b/internal/core/attribute.cpp
@@ -39,6 +39,16 @@ float Max::compute(
     return *std::max_element(segment.begin(), segment.end());
 }
 
+float MaxAt::compute(
+    ResampledSegment const & segment
+) noexcept (false) {
+    auto max_index = std::distance(
+        segment.begin(),
+        std::max_element(segment.begin(), segment.end())
+    );
+    return segment.sample_position_at(max_index);
+}
+
 float MaxAbs::compute(
     ResampledSegment const & segment
 ) noexcept (false) {
--- a/internal/core/attribute.cpp
+++ b/internal/core/attribute.cpp
@@ -24,6 +24,15 @@ float Min::compute(
     return *std::min_element(segment.begin(), segment.end());
 }
 
+float MinAt::compute(
+    ResampledSegment const & segment) noexcept(false) {
+    auto min_index = std::distance(
+            segment.begin(),
+            std::min_element(segment.begin(), segment.end())
+        );
+    return segment.sample_position_at(min_index);
+}
+
 float Max::compute(
     ResampledSegment const & segment
 ) noexcept (false) {
--- a/internal/core/attribute.cpp
+++ b/internal/core/attribute.cpp
@@ -49,15 +49,42 @@ float MaxAt::compute(
     return segment.sample_position_at(max_index);
 }
 
-float MaxAbs::compute(
+namespace {
+
+double max_abs(
     ResampledSegment const & segment
-) noexcept (false) {
+){
     auto max = *std::max_element(segment.begin(), segment.end(),
     [](const double& a, const double& b) { 
             return std::abs(a) < std::abs(b); 
         }
     );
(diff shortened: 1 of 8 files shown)


AdilKhurshid:
I have completely changed the implementation of the code compared to the way I did it before. Even I was not happy with the code that I wrote earlier 🙈. Thanks to your recent changes 😍to the code, which made the task much easier for me 😇. Please review these changes and let me know if there is anything that needs to be modified.

achaikou (approved):
Very nice! :sun_with_face: 
I love how careful you are with keeping the code pretty!

Beside one small comment, can't find anything else to complain about :cat:  I thought that maybe we can make implementation of maxabs_at a bit faster by using other logic, but nope, it has inconsistent performance, so your implementation wins!

So go merge, or if you like change this one small thing first :slightly_smiling_face:

internal/core/attribute.cpp
```diff
@@ -41,6 +60,26 @@ float MaxAbs::compute(
...
 
+float MaxAbsAt::compute(
+    ResampledSegment const & segment
+) noexcept (false) {
+
+    auto max_abs_val = *std::max_element(segment.begin(), segment.end(),
+    [](const double& a, const double& b) {
+            return std::abs(a) < std::abs(b);
+        }
+    );
```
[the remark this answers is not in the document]
AdilKhurshid: I agree that repeating code is not ideal, so I'm creating a new function like variance. Good Idea 😊

internal/core/attribute.cpp
```diff
@@ -49,9 +49,11 @@ float MaxAt::compute(
     return segment.sample_position_at(max_index);
 }
 
-float MaxAbs::compute(
+namespace {
+
+double maxAbs(
```
achaikou, line 54: Note that in this c++ file for function names we follow snake naming convention, so a more valid name for a function would have been max_abs or maxabs

internal/core/attribute.cpp
```diff
@@ -60,22 +62,27 @@ float MaxAbs::compute(
...
-            return std::abs(a) < std::abs(b);
-        }
-    );
+    auto max_abs_val = maxAbs(segment);
     auto max_abs_index = std::distance(
         segment.begin(),
         std::find_if(segment.begin(), segment.end(),
-            [&](const double& val) {
+        [&](const double& val) {
                 return std::abs(val) == std::abs(max_abs_val);
```
achaikou, line 82: Do you need to take std:abs from max_abs_val. Don't you already return std::abs ?

internal/core/attribute.cpp
AdilKhurshid: done 
https://github.com/equinor/vds-slice/pull/214/commits/15e91b991d3ddd90e96058fbbb5769827b1c2f08#diff-0a24dd0852852a74dc874fdfcbb7826193dfa983b4b02f34f87d06fb28e91a40R54

internal/core/attribute.cpp
AdilKhurshid: yes, fixed
https://github.com/equinor/vds-slice/commit/15e91b991d3ddd90e96058fbbb5769827b1c2f08#diff-0a24dd0852852a74dc874fdfcbb7826193dfa983b4b02f34f87d06fb28e91a40R82

achaikou (approved):
:+1: 
Looks good! Go merge!

PR #220, opened 2023-10-24
Add copy index.html to Dockerfile

yngve793:
Add single copy line to Dockerfile.


achaikou (approved):
Yep, looks like it is fixed!

Approving, but note that I still complain about commit names.
`Copy index.html`? where? why?

It would make more sense to name it
`Copy index.html to final Docker image`

And add a comment that we forgot to add it before and thus root page was anavailable.

`Set /server as working dir`

Also might benefit from adding "in final Docker image", but this at least implicitly defines it already.

But it would also benefit from comment that it makes more sense to have working dir as our app than whatever directory our base images decides to for us.

achaikou (approved):

PR #221, opened 2023-10-30
Add proper error message for invalid API requests

### LINKED ISSUE #208: Better error message on empty Get Request
Currently if one calls one of our endpoints without providing "query" parameter, user gets back
`{"error":"unexpected end of JSON input"}`

Reply is not clear enough, users would benefit from information that "query" parameter is expected.

AdilKhurshid:
closes #208

### DIFF
--- a/api/endpoint.go
+++ b/api/endpoint.go
@@ -276,8 +276,17 @@ func (request AttributeBetweenSurfacesRequest) execute(
 }
 
 func parseGetRequest(ctx *gin.Context, v Normalizable) error {
-	if err := json.Unmarshal([]byte(ctx.Query("query")), v); err != nil {
-		return core.NewInvalidArgument(err.Error())
+	query, status := ctx.GetQuery("query")
+	if (!status){
+		return core.NewInvalidArgument(
+			"GET request to specified endpoint requires a 'query' parameter",
+		)
+	}
+	if err := json.Unmarshal([]byte(query), v); err != nil {
+		msg := "Please ensure that the supplied query is valid " +
+			"and conforms to the expected swagger Request specification: %v"
+		return core.NewInvalidArgument(
+			fmt.Sprintf(msg, err.Error()))
 	}
 
 	if err := binding.Validator.ValidateStruct(v); err != nil {
--- a/cmd/query/main_test.go
+++ b/cmd/query/main_test.go
@@ -110,6 +110,16 @@ func TestSliceErrorHTTPResponse(t *testing.T) {
 		},
 		sliceTest{
 			baseTest{
+				name:           "GET request missing query parameter",
+				method:         http.MethodGet,
+				jsonRequest:    "MODIFIER: skip query parameter",
+				expectedStatus: http.StatusBadRequest,
+				expectedError:  "GET request to specified endpoint requires a 'query' parameter",
+			},
+			testSliceRequest{},
+		},
+		sliceTest{
+			baseTest{
 				name:           "Invalid json POST request",
 				method:         http.MethodPost,
 				jsonRequest:    "help I am a duck",
@@ -264,6 +274,16 @@ func TestFenceErrorHTTPResponse(t *testing.T) {
 				jsonRequest:    "help I am a duck",
 				expectedStatus: http.StatusBadRequest,
 				expectedError:  "invalid character",
+			},
+			testFenceRequest{},
+		},
+		fenceTest{
+			baseTest{
+				name:           "GET request missing query parameter",
+				method:         http.MethodGet,
+				jsonRequest:    "MODIFIER: skip query parameter",
+				expectedStatus: http.StatusBadRequest,
+				expectedError:  "GET request to specified endpoint requires a 'query' parameter",
 			},
 			testFenceRequest{},
 		},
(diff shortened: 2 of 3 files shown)


api/endpoint.go
```diff
@@ -277,7 +277,11 @@ func (request AttributeBetweenSurfacesRequest) execute(
 
 func parseGetRequest(ctx *gin.Context, v Normalizable) error {
 	if err := json.Unmarshal([]byte(ctx.Query("query")), v); err != nil {
-		return core.NewInvalidArgument(err.Error())
+		msg := "Your request could not be processed. " +
+			"Please ensure that the supplied data is valid " +
+			"and conforms to the expected swagger specification: "
+		return core.NewInvalidArgument(
```
achaikou, line 283: Can we have a test? :pleading_face: 

ErrorHTTPResponse tests in main_test.go are responsible for making sure our errors reach users and don't get lost along the way. Because of the duplication in the server code that we had they are repeated for every endpoint separately, but I am fine if you choose to stop repeating them for every endpoint and chose just one to test this case on.

achaikou (commented):
Complain for commit message: it now seems like we do not have proper error messages for any API requests. That's not true, thus confusing for whoever who would read the history.

We want to mention explicitly that the problem lies in not good enough error message when user is missing query parameter in GET requests.

cmd/query/main_test.go
```diff
@@ -108,6 +108,26 @@ func TestSliceErrorHTTPResponse(t *testing.T) {
...
+				method:         http.MethodGet,
+				jsonRequest:    "nil",
+				expectedStatus: http.StatusBadRequest,
+				expectedError:  "The specified endpoint requires a 'query' parameter ",
+			},
+			testSliceRequest{},
+		},
+		sliceTest{
+			baseTest{
+				name:           "Invalid json GET request",
```
achaikou, line 123: This test is actually the same as original one. What's the point in having two? So I think you can either:
1) Exchange original test with your test. It checks just presence of the error so it is fine. Original message is not really important
2) Add the full string to the test, thus combining both tests (would be fine, but the string is loong :smile: )
3) add to baseTest expectedErrorRegex and if it is present in the testcase compare with require.Regex. Then the string from 2 becomes shorter. Don't really see much value in this particular case, but can be as an option.

I think I would be fine with just option 1, but see if you prefer something else

cmd/query/main_test.go
```diff
@@ -108,6 +108,26 @@ func TestSliceErrorHTTPResponse(t *testing.T) {
 			},
 			testSliceRequest{},
 		},
+		sliceTest{
+			baseTest{
+				name:           "Invalid json GET request",
+				method:         http.MethodGet,
+				jsonRequest:    "nil",
```
achaikou, line 115: Maybe instead of nil identification use something like "MODIFIER: skip query parameter" so that whoever reads the tests has an idea that something special is about to happen in case of this value.

I am not 100% happy with it, but for now can't think of anything better. Your suggestion seems to be the best.

api/endpoint.go
```diff
@@ -276,8 +276,17 @@ func (request AttributeBetweenSurfacesRequest) execute(
...
+	if (!status){
+		return core.NewInvalidArgument(
+			"The specified endpoint requires a 'query' parameter  ",
+		)
+	}
+	if err := json.Unmarshal([]byte(query), v); err != nil {
+		msg := "Please ensure that the supplied query is valid " +
+			"and conforms to the expected swagger Request specification: "
+		return core.NewInvalidArgument(
+			fmt.Sprintf(msg+" %v", err.Error()))
```
achaikou, line 289: You can add %v at the end of your message instead. It would make it more clear that message expects some argument.

yngve793, line 289: Is the "fmt.Sprintf(msg+" %v", err.Error()))" for debugging or do we expect somebody to read the standard out on the server?

cmd/query/main_test.go
AdilKhurshid: I'm also agree with option one. So will change the test case

api/endpoint.go
AdilKhurshid: Yes, If an error occurs in this code, it will be logged on the server terminal just like any other error. In the production environment, we may want to hide this error from the server terminal. In that case, we need to modify the error handling in the code and add a proper logging mechanism for debugging in the production environment.

achaikou: Not only are those logged in to production server, they are forwarded to the user.
We also want to know where our users screw up! Helps with audit and problem-solving.

So these are proper production messages, no debugging.

yngve793: Ok, but I guess we want to turn off the output to terminal in the production release? Terminal messages are very slow and no-one will ever see the message.

achaikou: No, we do not want to turn them off.

They are very useful for us. By having them on output we ourselves can see what user did wrong, without relying on the client code (maybe they swallow the error message or handle it incorrectly). We can also spot weird behavior, aka someone trying to access unlisted storage accounts or constantly submitting wrong sas tokens.

And there is no point for special handling of this particular error message. Anyway we are logging all the requests on server console. If this is an error, I don't care about additional slowness of one more info.
Honest user errors won't happen that often.

achaikou (commented):
I am happy!
So as for me you can get ready for merge :cat:

yngve793 (approved):
No further comments.

PR #223, opened 2023-11-03
Abstract datahandle

yngve793:
This is a preliminary commit for the attribute difference(4D) endpoint.

The virtual differences cube is created by defining a virtual data source class used by all endpoints. Each new data source can now be created by implementing a new instance of the virtual class. This PR contains OvdsDataSource for accessing the *.vds files and OvdsMultiDataSource for data sources defined by a pair of *.vds files. Redefining the data source implies that current functions for attribute, fence and slice can be used without any changes. Thus, more code is reused and the the number of functions to tests is reduced. 

Remaining tasks:
- Test for new endpoints
- Test agains real data
- Load tests 
- Implement difference functions for fence and slice
- Verify output against real data
- ... Anything that is forgotten :-)


achaikou (commented):
Regarding your general idea: you were right, it's better! :+1:  Probably first two commits would have been already enough to convince me.

Regarding the code: I reviewed it as if it was a final merge-suggestion. Drafts in my mind look different :slightly_smiling_face: 
One of my complains here is that I think you introduced code that you don't really need (now or ever). So I recommend you read all my comments and if you agree with most major ones - start from scratch in new PR(s). Should be the cleanest choice.

yngve793 (commented):
This contains replay for comments up until the Go part of the code. The part that is addressed here updated and implemented in the DataSource interface PR. Remaining comments will be addressed in following PR´s.

yngve793:
Replaced by #231

PR #227, opened 2023-11-08
Adjust float values precision in C++ error strings

### LINKED ISSUE #193: Floating point precision in error messages
Many of our error messages include coordinates or other floating point values. When these are constructed in C++ these number by default includes 6 digits. In many cases that many digits is just noice. We should go over all errors and reduce the number of digits where it makes sense

AdilKhurshid:
When we are constructing error strings in C++, we often include coordinates or other floating point values. By default, these numbers include 6 digits, which can be unnecessary in many cases.

Therefore, setting the floating point precision to 2 decimal digits to eliminate any unnecessary noise.

closes #193

### DIFF
--- a/internal/core/cppapi_data.cpp
+++ b/internal/core/cppapi_data.cpp
@@ -17,6 +17,7 @@
 #include "regularsurface.hpp"
 #include "subcube.hpp"
 #include "subvolume.hpp"
+#include "utils.hpp"
 
 namespace {
 
@@ -203,7 +204,8 @@ void fence(
             if (!axis.inrange(coordinate[voxel])) {
                 if (fillValue == nullptr) {
                     const std::string coordinate_str =
-                        "(" +std::to_string(x) + "," + std::to_string(y) + ")";
+                        "(" +utils::to_string_with_precision(x, 6) + "," +
+                        utils::to_string_with_precision(y, 6) + ")";
                     throw detail::bad_request(
                         "Coordinate " + coordinate_str + " is out of boundaries "+
                         "in dimension "+ std::to_string(voxel)+ "."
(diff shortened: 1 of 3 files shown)


yngve793 (approved):
Looks good to me.

internal/core/cppapi_data.cpp
```diff
@@ -203,7 +204,7 @@ void fence(
             if (!axis.inrange(coordinate[voxel])) {
                 if (fillValue == nullptr) {
                     const std::string coordinate_str =
-                        "(" +std::to_string(x) + "," + std::to_string(y) + ")";
+                        "(" +utils::to_string_with_precision(x) + "," + utils::to_string_with_precision(y) + ")";
```
achaikou, line 207: This one is user-supplied data. Should we limit it as well? We will kinda limit it anyway by default 6 positions, but I am not sure how close the actual data is to each other - users generate the path themselves, likely in CDP coordinate system. Could it be that they usually differ at position 4 and 5? Idk...

From the other side I am not sure how this error message really helps users...

achaikou (commented):
From my side no complaints this time, just one question. I don't know the answer to it though and not sure who to ask (if we are going to bother).

internal/core/cppapi_data.cpp
AdilKhurshid: The precision level required varies depending on the seismic survey. I am unsure of the appropriate precision level needed for the error message. However, I can change it to 6 significant digits if that is acceptable.

achaikou: I am not the one to know if that's acceptable. In a perfect world one could just briefly discuss that with our PO as he should have experience with what is usual for surveys.
But what what I know 6 should be good enough here.
So go get ready for merge!

PR #229, opened 2023-11-15
Remove unnecessary e2e testcase

AdilKhurshid:
In our e2e test case, we  download the latest version of the blob library every time we run new container. However, the recent release of azure-storage-blob 12.19.0 has made it mandatory to include the permission argument in order to generate a SAS token.

Therefore, passing an empty parameter that would lead to an exception from the blob library and break code. So that testcase is no longer relevant and its good to remove.

[ref](https://github.com/Azure/azure-sdk-for-python/commit/86759fffbae2c93771607aacd2e6e5697b5b0bda)

### DIFF
--- a/tests/e2e/test_requests.py
+++ b/tests/e2e/test_requests.py
@@ -253,11 +253,6 @@ def test_attributes_between_surfaces():
             "409 Public access is not permitted",
             "401 Server failed to authenticate the request"
         ]
-    ),
-    (
-        generate_container_signature(
-            STORAGE_ACCOUNT_NAME, CONTAINER, STORAGE_ACCOUNT_KEY, permission=""),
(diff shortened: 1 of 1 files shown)


yngve793 (approved):
I agree that it make sense to remove the test.

achaikou (approved):
Only with comment that we are not removing it because they changed the API and we've been testing _it_, but because we just tested _using_ that API that one should have valid permissions to access our data.
But as this test was kinda just repeating Azure-itself functionality, it can be removed. It is not important.
Alternative would be to change permission to "w", but nah, removal is probably better.

AdilKhurshid:
Sure, I'm changing the commit message then

yngve793 (approved):
Appreciate better justification in the commit message.

PR #239, opened 2024-01-16
Add snyk to test docker image

### LINKED ISSUE #224: Scan docker image with Snyk and potentially include Snyk in GH actions

AdilKhurshid:
Added a GitHub action to scan the Docker image every time a pull request merges into the master branch. The results will be displayed in the GitHub action logs.

closes #224

### DIFF
--- a/.github/workflows/snyk.yaml
+++ b/.github/workflows/snyk.yaml
@@ -1,0 +1,42 @@
+name: Snyk test Docker image
+on: 
+  push:
+    branches: [master, radix]
+  workflow_dispatch:
+
+jobs:
+  snyk_test:
+    name: Snyk test
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Load openvds image
+      uses: "./.github/actions/load_openvds_image"
+
+    - name: Set docker image tag name for snyk testing
+      id: set_tag
+      run: |
+        tag=$(echo "${{ github.repository }}/docker-image-${{ github.ref_name }}" \
+         | tr '[:upper:]' '[:lower:]')
+        echo "DOCKER_IMAGE_TAG=$tag" >> "$GITHUB_ENV"
+
+    - name: Build docker image
+      run: |
+        docker build \
+        -f Dockerfile \
+        --build-arg OPENVDS_IMAGE=${{ env.OPENVDS_IMAGE_TAG }} \
+        --tag $DOCKER_IMAGE_TAG \
+        .
+
+    - name: Run Snyk to check Docker image for vulnerabilities
+      id: docker-image-scan
+      uses: snyk/actions/docker@master
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+      with:
+        command: monitor
+        image: $DOCKER_IMAGE_TAG
+        args:
+         --file=Dockerfile
+         --severity-threshold=high


.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Load openvds image
+      uses: "./.github/actions/load_openvds_image"
+
+    - name: Build docker image
+      run: |
+        tag=snyk_server
```
achaikou, line 19: Why is our image for testing named snyk server?
It shows up in the logs and is confusing

> Testing snyk_server...

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+    - name: Run Snyk to check Docker image for vulnerabilities
+      id: docker-image-scan
+      continue-on-error: true
+      uses: snyk/actions/docker@master
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+      with:
+        image: snyk_server
+        args: --file=Dockerfile
```
achaikou, line 34: Move to its own line? :innocent:

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,36 @@
...
+        --build-arg OPENVDS_IMAGE=${{ env.OPENVDS_IMAGE_TAG }} \
+        --tag $tag \
+        .
+
+    - name: Run Snyk to check Docker image for vulnerabilities
+      id: docker-image-scan
+      continue-on-error: true
+      uses: snyk/actions/docker@master
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
```
achaikou, line 31: Should we btw be using team default snyk token or should it be service account one? Were there any recommendations about it in the appsec group?

.github/workflows/snyk.yaml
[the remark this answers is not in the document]
AdilKhurshid: Good Soultion! We can set up this action to run as a daily scheduled job or upon someone creating a pull request to scan the vulnerabilities in the Docker image. If there are any high vulnerabilities detected, we can prioritize fixing them.

The real problem here we're facing is the absence of a good and reliable alert system that can promptly inform us of any security vulnerabilities found in docker image. Although I tried to integrate this with Slack alert for this purpose, it has been challenging to use in our strict ecosystem. As a result, someone from the team have to manually check the logs to identify any vulnerabilities.

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,36 @@
...
+        tag=snyk_server
+        docker build \
+        -f Dockerfile \
+        --build-arg OPENVDS_IMAGE=${{ env.OPENVDS_IMAGE_TAG }} \
+        --tag $tag \
+        .
+
+    - name: Run Snyk to check Docker image for vulnerabilities
+      id: docker-image-scan
+      continue-on-error: true
```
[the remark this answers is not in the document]
AdilKhurshid: Yes, I intentionally set this flag. If we don't set it, the action will fail even if it finds a low-level vulnerability, and there is no way to set it to only fail for high-level vulnerabilities, which is extremely annoying.

.github/workflows/snyk.yaml
AdilKhurshid: we can change the name. I just added this tag to distinguish the image name.

.github/workflows/snyk.yaml
AdilKhurshid: There is no proper recommendation, but we need a token for authentication from the Snyk account. Also, its only works with user account token so a security champion from our team can use its token.

.github/workflows/snyk.yaml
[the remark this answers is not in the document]
ajaust: @AdilKhurshid  Did you try the Action using with `command: monitor` as described in the [Action's README](https://github.com/snyk/actions)? It should add the container to our Snyk organization which should be somehow checked more regularly. It should also give warnings when a new vulnerability is found via Email or the Snyk bot.

The main question for me is: Would the GitHub Action also fails if we use the `monitor` command and the current image is has high or severe vulnerabilities. If not, we might have to run the normal test and `monitor`.

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,36 @@
+name: Snyk test Docker image
+on: 
+  push:
+    branches: [master]
```
ajaust, line 4: Maybe more a question to the more experienced developers: Should we also monitor the `radix` branch or inside of Radix good enough? I know that Radix also runs Snyk on the deployed image, but will they inform us if there is a vulnerability?

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,36 @@
...
+  workflow_dispatch:
+
+jobs:
+  snyk_test:
+    name: Snyk test
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
```
ajaust, line 12: Is there a reason to use `v3` over `v4`? I would suggest to use the newest supported version for new workflows.

.github/workflows/snyk.yaml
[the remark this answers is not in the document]
ajaust: I think we could overwrite the default arguments of `snyk test` using the `args` argument of the Action and set a severity threshold, see [Snyk docs](https://docs.snyk.io/snyk-cli/scan-and-maintain-projects-using-the-cli/set-severity-thresholds-for-cli-tests).

I just saw that you set the severity threshold to high? Isn't this enough to let the Action pass successfully if only low severity issues are found?

.github/workflows/snyk.yaml
achaikou: On Radix as far as I see it only shows a "bug" in UI near old and buggy deployment. We don't get any additional notifications as far as I see.

If our security champions think we want a faster reaction to this - then it is good idea to monitor at least Radix branch (test could stay whatever?)

.github/workflows/snyk.yaml
achaikou: I created service account when testing myself on my own fork... Seemed fine.
Btw, as far as I understand we are using team token? I don't know what value is there in snyk token variable, but I assumed it would be the team one.

.github/workflows/snyk.yaml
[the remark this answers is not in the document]
ajaust: From a short discussion on Slack it seems to be advisable to monitor all images that are deployed. Therefure, it would make sense to monitor the image that is used/created by the `radix` branch.

AdilKhurshid:
I think I have included all the changes that you guys mentioned above, and we discussed earlier. Please review this pull request again and let me know if any further changes are needed. Special thanks to @ajaust for helping me with this PR. You are a lifesaver 😊😇

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Load openvds image
+      uses: "./.github/actions/load_openvds_image"
+
+    - name: Set docker image tag name for snyk testing
+      id: set_tag
+      run: |
+        tag=$(echo "snyk/${{ github.repository }}/${GITHUB_REF#refs/heads/}" \
```
ajaust, line 20: I think it could make sense to change
`snyk/${{ github.repository }}/${GITHUB_REF#refs/heads/}`
to something like
`${{ github.repository }}/docker-image-${ github.ref_name }` because Snyk uses the tag as name in the dashboard.

To me this would have the following advantages:
- It is ok to leave out the `snyk/` prefix since it does not add any useful information in the Snyk dashboard (we are on Snyk so we know that it is related to Snyk.
- This should ensure that the scanned images get the same prefix, i.e., `equinor/vds-slice`, in the Snyk dashboard as the other `vds-slice` related scans.
- Makes it more clear in the Snyk dashboard what the scan result is for (the Docker image).

ajaust (commented):
It looks good overall. I agree with Alena's comments and I added another suggestions regarding the naming of the tag.

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,46 @@
...
+         | tr '[:upper:]' '[:lower:]')
+        echo "TAG=$tag" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
AdilKhurshid: Sure! I will add the above-mentioned changes. Thanks 😊

.github/workflows/snyk.yaml
AdilKhurshid: Good suggestion! 😇

ajaust (approved):
Looks good to me. 🙂

achaikou (approved):
Yes, looks very nice! :tada:  Go merge!

PR #245, opened 2024-01-29
Split core.go file into logical subparts

### LINKED ISSUE #215: Split core.go into logical subparts
The file internal/core/core.go is getting close to 1k lines and contains several types of function definitions. The length and combination of functions makes it harder than it has to be to find the dependencies between functions. 
Candidates for separations can be: 
- Function calls directly related to API endpoint requests. 
- Attribute related functions
- Parallelisation of the attribute request
- More simple support functions

Currently it is unclear how the import of the c code bundle these functions.

AdilKhurshid:
Splitting the primary core.go file into more logical sub-sections which will facilitate in easier code maintenance and readability in the future.

Closes #215


internal/core/core_fence.go
```diff
@@ -0,0 +1,79 @@
+package core
+
+/*
+#cgo LDFLAGS: -lopenvds
+#cgo CXXFLAGS: -std=c++17
+#include <capi.h>
+#include <ctypes.h>
+#include <stdlib.h>
```
achaikou, line 8: Didn't think that we would have to include all these in every file :pouting_cat: 

Wondering what is better - leave it as is or always include only capi.h, as it in itself actually has import of ctypes and stdlib. Probably would depend on compiler though if it compiles or not.

So I would probably vote for what you currently have, but would gladly hear other opinions :)

internal/core/core_fence.go
```diff
@@ -0,0 +1,79 @@
+package core
+
+/*
+#cgo LDFLAGS: -lopenvds
+#cgo CXXFLAGS: -std=c++17
```
achaikou, line 5: I think we cam get rid of these in slice, fence and attributes files and leave them in core.go one only? At least it still compiles if I remove these two lines.

internal/core/core.go
```diff
@@ -227,59 +226,6 @@ func GetInterpolationMethod(interpolation string) (int, error) {
 	}
 }
 
-func GetAttributeType(attribute string) (int, error) {
```
achaikou, line 230: Maybe we should leave it in core? This is the only one "translation" method that gets moved out. Not regular surface or bounds.

So maybe it would be easier to find here?

internal/core/core.go
yngve793, line 1: Can we move "type RegularSurface struct" and "cRegularSurface" to core_attributes.go? 
It seams like they are only used in this file.

yngve793, line 1: Specific metadata types like SliceMetadata can probably be moved to corresponding file.

yngve793 (commented):
Try to keep shared code in the remaining core.go.

internal/core/core_attributes.go
```diff
@@ -0,0 +1,454 @@
...
+		return nil, computeErrors[0]
+	}
+
+	out := make([][]byte, nAttributes)
+	for i := 0; i < nAttributes; i++ {
+		out[i] = buffer[i*mapsize : (i+1)*mapsize]
+	}
+
+	return out, nil
+}
```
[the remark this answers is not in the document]
AdilKhurshid: Sure

internal/core/core.go
AdilKhurshid: Sure

internal/core/core.go
AdilKhurshid: Yes, we can move those struct as well. Good suggestion

achaikou (approved):
I am happy :cat: As for me - rebase and merge, but lets wait for the second reviewer's stamp of approval and agreement :)

yngve793 (approved):
No further comments.

PR #254, opened 2024-02-16
Malloc

achaikou:
Ran performance tests. Most noticeable difference is for large-depth files.
For current master one performance test slice median goes down from 4.98s to 3.86s, for fence median goes down from 4.55 to 3.56.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 ARG OPENVDS_IMAGE=openvds
-ARG VDSSLICE_BASEIMAGE=golang:1.20-alpine3.16
+ARG VDSSLICE_BASEIMAGE=golang:1.20-alpine3.17
 FROM ${VDSSLICE_BASEIMAGE} as openvds
 RUN apk --no-cache add \
     curl \
@@ -88,7 +88,8 @@ RUN apk --no-cache add \
     libcurl \
     libxml2 \
     libuuid \
-    boost-log
+    boost-log \
+    jemalloc-dev
 
 WORKDIR /server
 COPY --from=installer /open-vds/Dist/OpenVDS/lib/* /open-vds/
@@ -101,4 +102,5 @@ USER 1001
 
 ENV LD_LIBRARY_PATH=/open-vds:$LD_LIBRARY_PATH
 ENV OPENVDS_AZURESDKFORCPP=1
+ENV LD_PRELOAD=/usr/lib/libjemalloc.so
 ENTRYPOINT [ "/server/query" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,12 +12,13 @@ RUN apk --no-cache add \
     boost-dev \
     libxml2-dev \
     libuv-dev \
-    util-linux-dev
+    util-linux-dev \
+    perl
 
 WORKDIR /
 RUN git clone https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds.git
 WORKDIR /open-vds
-RUN git checkout cbcd7b6163768118805dbcd080a5b0e386b82a6a
+RUN git checkout 3.3.3
 
 RUN cmake -S . \
     -B build \
--- a/internal/core/capi.cpp
+++ b/internal/core/capi.cpp
@@ -1,3 +1,5 @@
+#include <functional>
+
 #include "ctypes.h"
 #include "capi.h"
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -12,12 +12,13 @@ RUN apk --no-cache add \
...
     libuv-dev \
-    util-linux-dev
+    util-linux-dev \
+    perl
 
 WORKDIR /
 RUN git clone https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds.git
 WORKDIR /open-vds
-RUN git checkout cbcd7b6163768118805dbcd080a5b0e386b82a6a
+RUN git checkout 3.3.3
```
ajaust, line 21: 😻 I like that we can finally upgrade to a newer OpenVDS version.

Dockerfile
```diff
@@ -101,4 +102,5 @@ USER 1001
 
 ENV LD_LIBRARY_PATH=/open-vds:$LD_LIBRARY_PATH
 ENV OPENVDS_AZURESDKFORCPP=1
+ENV LD_PRELOAD=/usr/lib/libjemalloc.so
```
ajaust, line 105: Depending on whether we consider using `LD_PRELOAD` a fix or a workaround, we should consider opening an issue to keep track of the memory allocator. This would make it easier to revisit this once/of OpenVDS implements a different memory allocation strategy or helps us decide if we want to implement the allocator in a different way, e.g., linking OpenVDS against it.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG OPENVDS_IMAGE=openvds
-ARG VDSSLICE_BASEIMAGE=golang:1.20-alpine3.16
+ARG VDSSLICE_BASEIMAGE=golang:1.20-alpine3.17
```
ajaust, line 2: Are you happy with the bump to Alpine 3.17 or would you consider to bump even further, e.g., 3.19?

internal/core/capi.cpp
```diff
@@ -1,3 +1,5 @@
+#include <functional>
```
ajaust, line 1: Why is the header needed even though nothing has change in the C++ source files?

ajaust (approved):
It looks good to me. I that like that we can update to newer versions for some of our packages. I was also able to build the Docker image locally. 

I added some remarks for clarification and due to curiosity.

Dockerfile
ajaust: I also just realised that Go 1.22 is out which means that Go 1.20 is out of support. We could consider to upgrade to Go 1.22. It may give some better performance as well, see [Go 1.22 release notes](https://go.dev/doc/go1.22). I will try to build a Docker image with Go 1.22 locally and see if that shows any immediate blockers.

Dockerfile
```diff
@@ -12,12 +12,13 @@ RUN apk --no-cache add \
...
     libxml2-dev \
     libuv-dev \
-    util-linux-dev
+    util-linux-dev \
+    perl
 
 WORKDIR /
 RUN git clone https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds.git
 WORKDIR /open-vds
-RUN git checkout cbcd7b6163768118805dbcd080a5b0e386b82a6a
```
yngve793, line 20: Excellent that we finally can update to a proper release. 

I guess that commit "cbcd7b6163768118805dbcd080a5b0e386b82a6a" of open-vds was used for a specific reason. At least it was not trivial to update open-vds earlier. What is the reason and why is it okay now?

Dockerfile
```diff
@@ -12,12 +12,13 @@ RUN apk --no-cache add \
     boost-dev \
     libxml2-dev \
     libuv-dev \
-    util-linux-dev
+    util-linux-dev \
+    perl
```
yngve793, line 16: Where does the perl requirement come from? Is this a consequence of package upgrade or do we have a new dependency?

Dockerfile
```diff
@@ -100,4 +102,5 @@ USER 1001
 
 ENV LD_LIBRARY_PATH=/open-vds:$LD_LIBRARY_PATH
 ENV OPENVDS_AZURESDKFORCPP=1
+ENV LD_PRELOAD=/usr/lib/libjemalloc.so
```
yngve793, line 105: Is this file independent of architecture? In other words, is it compiled or does it belongs to the OS?

yngve793 (approved):
All tests are passing. 
It is very nice that we can use an updated version of open-vds. 
I have not verified the speed up.

Dockerfile
yngve793: Ok, thanks for the explanation.

PR #255, opened 2024-02-16
Prepare radixconfig for production

achaikou:
Some production app is partly created from test branches containing this PR code, but it will be set up properly during/after this PR merge.

Note: We can't have 0 replicas. If we set it like so, it is impossible to start the component. So it seems like we need to push to test environments only when we need to and stop them explicitly once we don't need them.
Alternative is to talk to Radix team and ask them to implement containers in stopped state.

### DIFF
--- a/radixconfig.yaml
+++ b/radixconfig.yaml
@@ -1,15 +1,19 @@
+#radixconfig production
 apiVersion: radix.equinor.com/v1
 kind: RadixApplication
 metadata:
-  name: oneseismictest
+  name: oneseismic
 spec:
   environments:
-    - name: dev
+    # stable production branch users are supposed to use
+    - name: prod
       build:
         from: radix
+    # environment that can be used for performance testing
+    # environment must be stopped when not needed
     - name: test
       build:
-        from: test
+        from: radix_test
   components:
     - name: server
       dockerfileName: Dockerfile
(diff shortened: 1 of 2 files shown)


radixconfig_playground.yaml
```diff
@@ -0,0 +1,46 @@
+#radixconfig_playground
```
ajaust, line 1: Is this comment line needed? It seems to duplicate the filename.

radixconfig.yaml
```diff
@@ -1,15 +1,19 @@
...
+  name: oneseismic
 spec:
   environments:
-    - name: dev
+    # stable production branch users are supposed to use
+    - name: prod
       build:
         from: radix
+    # environment that can be used for performance testing
+    # environment must be stopped when not needed
```
ajaust, line 13: You commented on this, but could it be possible to set the number of replicas to zero to avoid the environment from starting and then [scale the environment up via the CLI](https://radix.equinor.com/docs/topic-radix-cli/#examples) if needed? 

I have not used the Radix CLI yet, but it seems to be quite powerful.

ajaust (commented):
Looks good overall. I left a few comments with questions.

ajaust (approved):
Looks good. :)

PR #258, opened 2024-03-07
Visualize performance tests results

achaikou:
Sends data to the Azure Monitor service via local prometheus.

What tests we want to run, what configurations we need, what we want to show in grafana is a different topic, but if you have any immediate improvement suggestions - comment straight away.

`Performance tests` workflow would be configured the way so every run over provided ENDPOINT secret sends data to the azure monitor instance. We'd need to decide whether we want to run that on schedule on not, also against which endpoint.
Currently runs can be found on my fork and tests results should be visible in grafana.

`Performance (manual)` workflow doesn't send anything anywhere.

Resources (Azure Monitor and Grafana) can be found in our Azure test resource group with prefix performance.

### DIFF
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -92,6 +92,7 @@ jobs:
 
       - name: Run "${{ matrix.name }}" test
         env:
+          TEST_NAME: ${{ matrix.name }}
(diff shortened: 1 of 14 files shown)


ajaust (approved):
Awesome! Looks good. Feel free to squash the fixups and merge. 🎉

PR #259, opened 2024-03-11
Expose storage account in metric

achaikou:
Playground grafana is updated with respective panel, production grafana will be changed after merge (might need to discuss the need to exclude test storage account from all the metrics)

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -49,7 +49,7 @@ func NewMetrics() *Metrics {
 		}, []string{"path", "status"}),
 
 		requestCount: prometheus.NewCounterVec(prometheus.CounterOpts{
-			Name: "vdsslice_number_of_requests",
+			Name: "vdsslice_requests_count",
 			Help: "VDSslice number of requests.",
 		}, []string{"method", "path"}),
 	}
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -1,9 +1,10 @@
 package metrics
 
-
 import (
-	"time"
+	"net/url"
+	"slices"
 	"strconv"
+	"time"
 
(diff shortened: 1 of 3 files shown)


internal/metrics/metrics.go
```diff
@@ -82,11 +84,35 @@ func NewGinMiddleware(metrics *Metrics) gin.HandlerFunc {
...
+	for _, vdsURL := range vdsURLs {
+		parsedURL, err := url.Parse(vdsURL)
+		if err != nil {
+			storageAccounts = append(storageAccounts, vdsURL)
+			continue
+		}
+		storageAccounts = append(storageAccounts, parsedURL.Hostname())
+	}
+
+	slices.Sort(storageAccounts)
```
yngve793, line 103: Why are these sorted? Is the user expect these to be returned in alphabetical order?

yngve793 (commented):
Looks good. 
Are these storages accounts only used for monitoring. 
Can there be duplicates and if so is it desired to have multiple instances of same storage account.

internal/metrics/metrics.go
achaikou: Because apparently by sorting and using compact one can remove duplicates using slice package.

achaikou (commented):
This metric is exposed for monitoring only.

Not sure about another question, but I decided to remove duplicates from the requests (double vds source case) as we are interested in how many requests go to our storage accounts and how many users use their own.
Maybe actually should have counted it 2 times, but decided that it is more likely that both vds would go to the same storage account, so lets count it once not twice.

internal/metrics/metrics.go
yngve793: Ok, fine.

yngve793 (approved):
Tests passes and code looks good.

Issue #78, opened 2022-12-22
Un-pin the version of the go alpine base image in Docker


ErlendHaa:
We are currently pinned on an older alpine due to OpenVDS' build process breaking  on some dependancy. OpenVDS has fixed this upstream, but as of now the fix is not part of any release tag. Once the next OpenVDS release is out (v3.1.3) we should upgrade to that version and unpin the go alpine image again. (See #70)

ajaust:
Our current image is based on Alpine 3.16 which reaches its end of life on 23 May 2024. This is not too far in the future so we should keep an eye on this. We either should use a different base image (Ubuntu) or try to upgrade to a newer version of Alpine. Using a newer version of Alpine will require to upgrade OpenVDS or some patching of OpenVDS.

PR #263, opened 2024-04-08
Update OpenVDS version to 3.4.1

ajaust:
This update finally allows us to unpin the Alpine version used in the Docker image.

This PR closes https://github.com/equinor/vds-slice/issues/78.


yngve793 (approved):
Builds and works fine.

PR #264, opened 2024-04-11
Interpolation near border

### LINKED ISSUE #252: What to do with border points stolen for interpolation in attributes endpoints
We calculate attributes on trace subset which can be very close to the trace border points.
As requested user values do not directly fall onto survey samples, we use interpolation algorithms to resample the data for users needs.

Near the end of the trace we are however lacking enough samples to perform good interpolation. Idea of implementing extrapolation on the borders was discarded as less precise solution which needs more efforts.
Instead we went on with "stealing" couple of samples from survey for interpolation purposes. Assumption was that users are not that interested in the data near surface borders and will hardly ever notice that. This assumption might have been wrong.

Current number of stolen samples from each border is 2. This number was set by looking closer at different interpolation algorithms. However it was supposed to be internal value and subject to a change if we ever looked at interpolation logic again and with more knowledge.

Aka if survey has samples `100, 105, 110, 115, 120, 125, 130`, then if user requests range `[117, 127]`, we would throw because we don't have two samples to the other side from `127`. Actually the same would happen for user input `[117, 125]` as `125` also doesn't have 2 samples on its right and we for simplicity ignore the fact that `125` is on the sample itself.

Problem here is that this information is hidden from the end-user. User is sure that by sending borders of `[117, 127]` their request should be handled, when it is not.

How can we deal with it?

1. If we update documentation with this information, we might lock user scripts to always rely on 2 samples, thus making internal information public and making it more difficult to change.
2. If we update error message with actual boundaries we face similar problem as users would still need to adjust to the error somehow, only without knowing how and why. Also at the time error message is thrown original context of user request is long stripped off, thus making it difficult to tell user what they did wrong.
3. We might want to revisit interpolation logic. Might it be that 1 sample on the side is good enough (aka distortion provided by having just 1 sample instead of 2 is not that big)? And if user range ends up with a sample, we shouldn't need any additional samples at all? In that case we might drop this problem for the user altogether. That however might require more deep understanding of interpolation than we currently have.
4. SC Day suggestion: use linear interpolation at the very end. Pros: it is likely to make more sense than whatever cubic interpolation comes up with in the end. Cons: we need to adjust to it at the stage of retrieving the samples, which might mean we need to pass down the flag that this one has less samples than expected.

### LINKED ISSUE #253: Adjust above/below documentation for attributes
It is now not clear that values that are in above/below range should be inbounds.

Currently we have just "Samples that are out-of-range of the seismic volume in the vertical plane are considered an error." written in the main text and it is not enough.

achaikou:
Closes #252 
Closes #253

### DIFF
--- a/internal/core/core_attributes_test.go
+++ b/internal/core/core_attributes_test.go
@@ -67,54 +67,112 @@ func TestSurfaceWindowVerticalBounds(t *testing.T) {
 	testcases := []struct {
 		name     string
 		values   [][]float32
+		below    float32
+		above    float32
 		inbounds bool
 	}{
-		// 2 samples is the margin needed for interpolation
+		// top boundary at 4.0, bottom boundary at 40.0
 		{
-			name:     "Top boundary is on sample 2 samples away from first depth recording",
-			values:   [][]float32{{16.00}},
-			inbounds: true,
-		},
-		{
-			name:     "Top boundary snaps to sample 2 samples away from the top",
-			values:   [][]float32{{13.00}},
-			inbounds: true,
-		},
-		{
-			name:     "Top sample is less than 2 samples from the top",
+			name:     "value - above is 2 samples away from the first existing sample",
 			values:   [][]float32{{12.00}},
+			above:    0,
+			below:    12,
+			inbounds: true,
+		},
+		{
+			name:     "value - above snaps to sample which is 2 samples away from the first existing sample",
+			values:   [][]float32{{9.00}},
+			above:    0,
+			below:    12,
+			inbounds: true,
+		},
+		{
+			name:     "value - above is 1 sample away from the first existing sample",
+			values:   [][]float32{{8.00}},
+			above:    0,
+			below:    12,
+			inbounds: true,
+		},
+		{
+			name:     "value - above snaps to sample which is 1 sample away from the first existing sample",
+			values:   [][]float32{{7.00}},
+			above:    0,
+			below:    12,
+			inbounds: true,
+		},
+		{
+			name:     "value - above is on the first existing sample",
+			values:   [][]float32{{4.00}},
+			above:    0,
+			below:    12,
+			inbounds: true,
+		},
+		{
+			name:     "value - above is less than first existing sample",
+			values:   [][]float32{{3.00}},
+			above:    0,
+			below:    12,
 			inbounds: false,
 		},
 		{
-			name:     "Bottom boundary is on sample 2 samples away from last depth recording",
-			values:   [][]float32{{28.00}},
-			inbounds: true,
-		},
-		{
-			name:     "Bottom boundary snaps to sample 2 samples away from last depth recording",
-			values:   [][]float32{{31.00}},
-			inbounds: true,
-		},
-		{
-			name:     "Bottom sample is less than 2 samples from last depth recording",
+			name:     "value + below is 2 samples away from the last existing sample",
 			values:   [][]float32{{32.00}},
+			above:    12,
+			below:    0,
+			inbounds: true,
+		},
+		{
+			name:     "value + below snaps to sample which is 2 samples away from the last existing sample",
+			values:   [][]float32{{35.00}},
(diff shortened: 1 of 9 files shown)


tests/gtest/cppapi_test.cpp
```diff
@@ -264,6 +264,64 @@ TEST_F(SubvolumeTest, ManyFills)
...
+        7, 4,
+    };
+
+    std::array<float, size> bottom_surface_data = {
+        40, 39,
+        36, 35,
+        32, 32,
+    };
+
+    const std::size_t expected_fetched_size = 10;
```
yngve793, line 291: I do not understand the purpose of this test. 
Independently of the top and bottom surfaces the entire trance is returned. 
The check consist of verifying that all 10 sample positions are in all six traces?

yngve793 (approved):
All commits compiles and passes the tests. 
Most of the commits are straight forward and there is not much to comment on. 
Commit story is fine and the code passes various poking tests.

internal/core/core_attributes_test.go
```diff
@@ -67,54 +67,112 @@ func TestSurfaceWindowVerticalBounds(t *testing.T) {
...
-			values:   [][]float32{{22.00, 32.00, 12.00}, {18.00, 31.00, 28.00}, {16.00, 15.00, 11.00}},
+			values:   [][]float32{{22.00, 40.00, 8.00}, {41.00, 35.00, 32.00}, {12.00, 11.00, 7.00}},
+			above:    4,
+			below:    4,
 			inbounds: false,
 		},
 		{
 			name:     "Fillvalue should not be bounds checked",
 			values:   [][]float32{{-999.25}},
+			above:    4,
```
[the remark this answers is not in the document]
yngve793: What about something like:
name:     "For attributes Fillvalue means ignore trace and should not be bounds checked",

internal/core/subvolume.hpp
```diff
@@ -35,6 +35,15 @@ float ceil_with_tolerance(float x);
  */
 class SegmentBlueprint {
 public:
+    /**
+     * Calculates sample position at provided index.
+     *
+     * @param index
+     * @param zero_index_sample_position Sample position at index 0
+     */
+    float sample_position_at(int index, float zero_index_sample_position) const noexcept{
```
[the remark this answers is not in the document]
yngve793: As an user of the class I expect to be able to understand what to expect of the function without running test inputs and check the returned result. Either a doc string is used to explain input, output and the expected behaviour. Alternatively all this information must be represented in the function name and input variable names. I am not a fan of reading through the entire class in order to use a single function. Any solution where the user can figure out what the function expect and what is expected to be returned without reading the hole class is fine by me.

tests/gtest/cppapi_test.cpp
yngve793: Ok, then I understand. Returning all 10 samples is probably among the most likely results in case of a bug and only checking annotated position do not distinguish the different traces. 

What about checking top and bottom separately and stopping before the edge in the non checking direction.
All data values are already available in the file under the name 'samples_10_data'. Why don't we check the data value as well as the position?

yngve793 (commented):
Replay to comments.

tests/gtest/cppapi_test.cpp
yngve793: The code handle 'top' and 'bottom' separately. Given this it makes sense to have separate tests for the two cases. In case the tests fails in the future it is useful for the developer how made the change causing the failure to know if it was the top or the bottom that failed.

yngve793 (approved):
Closed some, commented on one.

tests/gtest/cppapi_test.cpp
achaikou: If you ask for top=4, reference=20, bottom=20, you get
4, 8, 12, 16, 20, 24, 28.

You still have no clue what really failed, you just know what was tested.

But fine, if that is important to you, lets split.

PR #265, opened 2024-04-15
Test expired sas token accessing cache

### LINKED ISSUE #241: Test reading Cache on expired sas token
We are currently storing previous computed results in cache. It would be useful to test the following after an initial request:
- Verify that a new request with valid SAS token is read from cache
- Verify that a new request with invalid SAS token is ignored and no data is returned
- Verify that a new request with no SAS token is ignored and no data is returned

achaikou:
Closes #241

### DIFF
--- a/tests/e2e/test_caching.py
+++ b/tests/e2e/test_caching.py
@@ -16,6 +16,23 @@ cached_data_paths_payloads = [
     ("attributes/surface/between", payload_merge(connection_payload([SAMPLES10_URL], sas=[DUMMY_SAS]), attributes_between_surfaces_payload())),
 ]
 
+
+def make_caching_call(payload, path, sas=None):
+    """
+    makes a payload call to path.
+    used as a payload caching call
+    """
+    if sas == None:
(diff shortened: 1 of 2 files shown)


tests/e2e/test_caching.py
```diff
@@ -66,3 +67,23 @@ def test_cached_data_access_with_various_sas(path, payload, token, status, error
...
+    expire_time = 2
+    sas = generate_container_signature(
+        STORAGE_ACCOUNT_NAME,
+        CONTAINER,
+        STORAGE_ACCOUNT_KEY,
+        lifespan=expire_time,
+        permission=blob.ContainerSasPermissions(read=True))
+
+    make_caching_call(payload, path, sas)
+    time.sleep(expire_time + 1)
```
yngve793, line 83: Very nice test. :-)

yngve793 (approved):
Refactoring of test code, OK. 
All tests passes, OK.
The test fails when there is time for the second call.  Very good.

PR #269, opened 2024-06-06
Dependency cleanup

achaikou:
Cleans up calls a bit with intention of reusing functionality where makes sense.
Adds few testcases which reveal some bugs.

### DIFF
--- a/internal/core/coordinate_transformer.hpp
+++ b/internal/core/coordinate_transformer.hpp
@@ -116,8 +116,8 @@ public:
         return m_transformer_a.WorldToAnnotation(worldPosition);
     }
     OpenVDS::DoubleVector3 IJKPositionToAnnotation(const OpenVDS::DoubleVector3& ijkPosition) const {
-        auto ijkPosistionInCubeA = as_cube_a_ijk_position(ijkPosition);
-        return m_transformer_a.IJKPositionToAnnotation(ijkPosistionInCubeA);
+        auto ijkPositionInCubeA = as_cube_a_ijk_position(ijkPosition);
+        return m_transformer_a.IJKPositionToAnnotation(ijkPositionInCubeA);
     }
 
     void to_cube_a_voxel_position(float* out_cube_a_position, float const* intersection_cube_position) const {
--- a/internal/core/coordinate_transformer.hpp
+++ b/internal/core/coordinate_transformer.hpp
@@ -120,7 +120,14 @@ public:
         return m_transformer_a.IJKPositionToAnnotation(ijkPositionInCubeA);
     }
 
-    void to_cube_a_voxel_position(float* out_cube_a_position, float const* intersection_cube_position) const {
+    /**
+     * Having position in intersection voxel coordinates, transforms it to cube
+     * a voxel coordinates. Only 3 dimensions are supported: first 3 T values
+     * will be read from intersection_cube_position and 3 T values will be
+     * written to out_cube_a_position.
+     */
+    template<typename T>
+    void to_cube_a_voxel_position(T* out_cube_a_position, T const* intersection_cube_position) const {
         for (int ijk_index = 0; ijk_index < 3; ++ijk_index) {
             auto voxel_index = m_transformer_a.IJKToVoxelDimensionMap()[ijk_index];
             out_cube_a_position[voxel_index] = intersection_cube_position[voxel_index] +
@@ -128,7 +135,14 @@ public:
         };
     }
 
-    void to_cube_b_voxel_position(float* out_cube_b_position, float const* intersection_cube_position) const {
+    /**
+     * Having position in intersection voxel coordinates, transforms it to cube
+     * b voxel coordinates. Only 3 dimensions are supported: first 3 T values
+     * will be read from intersection_cube_position and 3 T values will be
+     * written to out_cube_b_position.
+     */
+    template<typename T>
+    void to_cube_b_voxel_position(T* out_cube_b_position, T const* intersection_cube_position) const {
         for (int ijk_index = 0; ijk_index < 3; ++ijk_index) {
             auto voxel_index = m_transformer_a.IJKToVoxelDimensionMap()[ijk_index];
             out_cube_b_position[voxel_index] = intersection_cube_position[voxel_index] +
--- a/internal/core/coordinate_transformer.hpp
+++ b/internal/core/coordinate_transformer.hpp
@@ -45,6 +45,10 @@ public:
         return coordinate_transformer.WorldToAnnotation(worldPosition);
     }
 
+    OpenVDS::IntVector3 IJKToVoxelDimensionMap() const {
+        return coordinate_transformer.IJKToVoxelDimensionMap();
+    }
+
 private:
     OpenVDS::IJKCoordinateTransformer coordinate_transformer;
 };
@@ -56,6 +60,10 @@ public:
         SingleCoordinateTransformer const& transformer_b
     )
         : m_transformer_a(transformer_a) {
+
+        if (transformer_a.IJKToVoxelDimensionMap() != transformer_b.IJKToVoxelDimensionMap()) {
+            throw std::runtime_error("Coordinate Transformers have different dimension maps");
+        }
 
         /*
          * For each dimension, intersection 0 index corresponds to either 0-line
@@ -89,7 +97,7 @@ public:
 
     OpenVDS::IntVector3 VoxelIndexToIJKIndex(const OpenVDS::IntVector3& voxelIndex) const {
         // voxel index to ijk index depends only on dimensions order, which
-        // should be the same for intersection and cube a
+        // should be the same for all the cubes
         return m_transformer_a.VoxelIndexToIJKIndex(voxelIndex);
     }
     OpenVDS::DoubleVector3 IJKIndexToWorld(const OpenVDS::IntVector3& ijkIndex) const {
@@ -112,15 +120,19 @@ public:
         return m_transformer_a.IJKPositionToAnnotation(ijkPosistionInCubeA);
     }
 
-    void to_cube_a_ijk_position(float* out_cube_a_position, float const* intersection_cube_position) const {
-        for (int index = 0; index < 3; ++index) {
-            out_cube_a_position[index] = intersection_cube_position[index] + this->m_intersection_zero_as_cube_a_index[index];
-        }
+    void to_cube_a_voxel_position(float* out_cube_a_position, float const* intersection_cube_position) const {
+        for (int ijk_index = 0; ijk_index < 3; ++ijk_index) {
+            auto voxel_index = m_transformer_a.IJKToVoxelDimensionMap()[ijk_index];
+            out_cube_a_position[voxel_index] = intersection_cube_position[voxel_index] +
+                                               this->m_intersection_zero_as_cube_a_index[ijk_index];
+        };
     }
 
-    void to_cube_b_ijk_position(float* out_cube_b_position, float const* intersection_cube_position) const {
-        for (int index = 0; index < 3; ++index) {
-            out_cube_b_position[index] = intersection_cube_position[index] + this->m_intersection_zero_as_cube_b_index[index];
+    void to_cube_b_voxel_position(float* out_cube_b_position, float const* intersection_cube_position) const {
+        for (int ijk_index = 0; ijk_index < 3; ++ijk_index) {
+            auto voxel_index = m_transformer_a.IJKToVoxelDimensionMap()[ijk_index];
+            out_cube_b_position[voxel_index] = intersection_cube_position[voxel_index] +
+                                               this->m_intersection_zero_as_cube_b_index[ijk_index];
         }
     }
 
--- a/tests/gtest/cppapi_test.cpp
+++ b/tests/gtest/cppapi_test.cpp
@@ -43,15 +43,9 @@ std::map<Samples10Points, std::vector<float>> samples_10_data(float fill)
 
 class SubvolumeTest : public ::testing::Test {
 protected:
-    void SetUp() override {
-        datahandle = make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str());
-    }
+    SubvolumeTest() : datahandle(make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str())) {}
 
-    void TearDown() override {
-        delete datahandle;
-    }
-
-    SingleDataHandle* datahandle;
+    SingleDataHandle datahandle;
 
     static constexpr float fill = -999.25;
     std::map<float, std::vector<float>> points;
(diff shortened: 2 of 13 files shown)


tests/gtest/cppapi_test.cpp
```diff
@@ -43,15 +43,9 @@ std::map<Samples10Points, std::vector<float>> samples_10_data(float fill)
...
 class SubvolumeTest : public ::testing::Test {
 protected:
-    void SetUp() override {
-        datahandle = make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str());
-    }
-
-    void TearDown() override {
-        delete datahandle;
-    }
+    SubvolumeTest() : datahandle(make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str())) {}
```
sveinung-r, line 46: This changes the behaviour of this class beyond just switching from a heap allocated SingleDataHandle. Now a new data handle is created at construction and destroyed at destruction of the SubVolumeTest class, rather than using the setup and teardown methods.

I have not looked into the details of how this class is used, so I don't know if this breaks anything. Was this intentional?

tests/gtest/cppapi_test.cpp
```diff
@@ -374,15 +368,9 @@ TEST_F(SubvolumeTest, LargestPossibleMarginRetrievedNearTraceBottomBoundary)
...
 class SurfaceAlignmentTest : public ::testing::Test {
 protected:
-    void SetUp() override {
-        datahandle = make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str());
-    }
-
-    void TearDown() override {
-        delete datahandle;
-    }
+    SurfaceAlignmentTest() : datahandle(make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str())) {}
```
sveinung-r, line 371: Same as above

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -24,6 +24,8 @@ const std::string CREDENTIALS = "";
 
 
 class DatahandleAttributeTest : public ::testing::Test {
+protected:
```
sveinung-r, line 27: Same as above

tests/gtest/datahandle_fence_test.cpp
```diff
@@ -19,6 +19,8 @@ const std::string SHIFT_8_BIG_DATA = "file://shift_8_32x3_cube.vds";
 const std::string CREDENTIALS = "";
 
 class DatahandleFenceTest : public ::testing::Test {
+protected:
+    DatahandleFenceTest() : single_datahandle(make_single_datahandle(REGULAR_DATA.c_str(), CREDENTIALS.c_str())) {}
```
sveinung-r: Same as above

tests/gtest/datahandle_metadata_test.cpp
```diff
@@ -17,6 +17,8 @@ const std::string SHIFT_8_BIG_DATA = "file://shift_8_32x3_cube.vds";
 const std::string CREDENTIALS = "";
 
 class DatahandleMetadataTest : public ::testing::Test {
+protected:
+    DatahandleMetadataTest() : single_datahandle(make_single_datahandle(REGULAR_DATA.c_str(), CREDENTIALS.c_str())) {}
```
sveinung-r: Here as well

tests/gtest/datahandle_slice_test.cpp
```diff
@@ -19,7 +19,8 @@ const std::string SHIFT_8_BIG_DATA = "file://shift_8_32x3_cube.vds";
 const std::string CREDENTIALS = "";
 
 class DatahandleSliceTest : public ::testing::Test {
-
+protected:
+    DatahandleSliceTest() : single_datahandle(make_single_datahandle(REGULAR_DATA.c_str(), CREDENTIALS.c_str())) {}
```
sveinung-r: And here

tests/gtest/cppapi_test.cpp
```diff
@@ -782,24 +782,17 @@ void inplace_subtraction(float* buffer_A, const float* buffer_B, std::size_t nsa
...
-            SAMPLES_10_x2.c_str(),
-            CREDENTIALS.c_str(),
-            binary_operator::SUBTRACTION
-        );
-    }
-
-    void TearDown() override {
-        delete double_datahandle;
-    }
+    FenceFunctionTest() : datahandle(make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str())),
```
sveinung-r, line 785: Same as for SingleDatahandle...

tests/gtest/cppapi_test.cpp
```diff
@@ -857,24 +850,17 @@ TEST_F(FenceFunctionTest, RequestingFenceDataSubtract) {
...
-            SAMPLES_10_x2.c_str(),
-            CREDENTIALS.c_str(),
-            binary_operator::SUBTRACTION
-        );
-    }
-
-    void TearDown() override {
-        delete double_datahandle;
-    }
+    SliceFunctionTest() : datahandle(make_single_datahandle(SAMPLES_10.c_str(), CREDENTIALS.c_str())),
```
sveinung-r, line 853: Here as well

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -25,7 +25,28 @@ const std::string CREDENTIALS = "";
 
 class DatahandleAttributeTest : public ::testing::Test {
 protected:
-    DatahandleAttributeTest() : single_datahandle(make_single_datahandle(REGULAR_DATA.c_str(), CREDENTIALS.c_str())) {}
+    DatahandleAttributeTest() : single_datahandle(make_single_datahandle(REGULAR_DATA.c_str(), CREDENTIALS.c_str())),
```
sveinung-r, line 28: And here

tests/gtest/datahandle_fence_test.cpp
```diff
@@ -20,7 +20,30 @@ const std::string CREDENTIALS = "";
 
 class DatahandleFenceTest : public ::testing::Test {
 protected:
-    DatahandleFenceTest() : single_datahandle(make_single_datahandle(REGULAR_DATA.c_str(), CREDENTIALS.c_str())) {}
+    DatahandleFenceTest() : single_datahandle(make_single_datahandle(REGULAR_DATA.c_str(), CREDENTIALS.c_str())),
```
sveinung-r, line 23: One more

tests/gtest/cppapi_test.cpp
achaikou: To all of your comments:
yes, it was intentional because otherwise there were issues with compilation (member initialization).

According to gtest [documentation](https://github.com/google/googletest/blob/main/docs/faq.md#should-i-use-the-constructordestructor-of-the-test-fixture-or-setupteardown-ctorvssetup) that is almost the same, so I didn't bother put it into spotlight.

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -91,43 +103,75 @@ class DatahandleAttributeTest : public ::testing::Test {
...
 
-        const double xori = cdp[0].first;
-        const double yori = cdp[0].second;
+        double xori = cdp[0].first;
+        double yori = cdp[0].second;
         const double xinc = std::hypot(iline_distance_x, iline_distance_y) / nsteps_iline;
         const double yinc = std::hypot(xline_distance_x, xline_distance_y) / nsteps_xline;
         const double rotation = std::atan2(iline_distance_y, iline_distance_x) * 180 / M_PI;
 
+        xori += iline_distance_x / nsteps_iline * origin_steps_shift_iline +
```
sveinung-r, line 112: Why change the order of the inline and line operands between the lines? Would be a lot easier to read this if it was the same.

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -233,4 +277,59 @@ TEST_F(DatahandleAttributeTest, Attribute_Different_Size_Double) {
...
 }
 
+TEST_F(DatahandleAttributeTest, Attribute_Double_Data_Outside_Intersection_Horizontal_Plane) {
+    DoubleDataHandle& datahandle = double_datahandle;
+    int ilines_before = 3;
+    int ilines_after = 1;
+    int xlines_before = 1;
+    int xlines_after = 2;
+
+    Grid grid = get_grid(datahandle, -ilines_before, -xlines_before);
```
sveinung-r, line 287: I don't understand why origin is shifted by the before il/xl values. Is it possible to clarify this with better naming? Or a comment if not.

sveinung-r: I think part of my confusion stems from my interpretation of before/after as temporal. As in this is the number of inlines before running the request and this is the number after.

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -233,4 +277,59 @@ TEST_F(DatahandleAttributeTest, Attribute_Different_Size_Double) {
     delete subvolume;
 }
 
+TEST_F(DatahandleAttributeTest, Attribute_Double_Data_Outside_Intersection_Horizontal_Plane) {
```
sveinung-r, line 280: I think this test needs a few comments (and perhaps a little sketch) to understand what is going on.

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -233,4 +277,59 @@ TEST_F(DatahandleAttributeTest, Attribute_Different_Size_Double) {
...
+    static std::vector<float> pri_surface_data(nrows * ncols, 36.0f);
+    static std::vector<float> bot_surface_data(nrows * ncols, 52.0f);
+    RegularSurface pri_surface = RegularSurface(pri_surface_data.data(), nrows, ncols, grid, fill);
+    RegularSurface top_surface = RegularSurface(top_surface_data.data(), nrows, ncols, grid, fill);
+    RegularSurface bot_surface = RegularSurface(bot_surface_data.data(), nrows, ncols, grid, fill);
+
+    EXPECT_THAT([&]() {
+        make_subvolume(datahandle.get_metadata(), pri_surface, top_surface, bot_surface);
+    },
+                testing::ThrowsMessage<std::runtime_error>(testing::HasSubstr("Vertical window is out of vertical bounds")));
```
sveinung-r, line 332: Very confusing indentation

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -233,4 +277,59 @@ TEST_F(DatahandleAttributeTest, Attribute_Different_Size_Double) {
...
+
+    int requested_low[3] = {expected_low[0] - ilines_before, expected_low[1] - xlines_before, expected_low[2]};
+    int requested_high[3] = {expected_high[0] + ilines_after, expected_high[1] + xlines_after, expected_high[2]};
+
+    check_attribute(*subvolume, requested_low, requested_high, expected_low, expected_high, 2);
+
+    delete subvolume;
+}
+
+TEST_F(DatahandleAttributeTest, Attribute_Double_Data_Outside_Intersection_Vertical_Plane) {
```
sveinung-r, line 314: I think it would be useful with a short explanation of what actually goes on in this test. I assume one of the surfaces within the bounds of one of the intersecting cubes, but outside the bounds of the intersection?

tests/gtest/datahandle_slice_test.cpp
```diff
@@ -93,10 +99,10 @@ class DatahandleSliceTest : public ::testing::Test {
                     int iline = value & 0xFF0000;
                     iline = iline >> 16;
                     xline = xline >> 8;
-
-                    EXPECT_EQ(iline, iline_array[il]);
-                    EXPECT_EQ(xline, xline_array[xl]);
-                    EXPECT_EQ(sample, sample_array[s]);
+;
```
sveinung-r, line 102: ???

internal/core/coordinate_transformer.hpp
```diff
@@ -112,15 +120,19 @@ class DoubleCoordinateTransformer : public CoordinateTransformer {
         return m_transformer_a.IJKPositionToAnnotation(ijkPosistionInCubeA);
     }
 
-    void to_cube_a_ijk_position(float* out_cube_a_position, float const* intersection_cube_position) const {
-        for (int index = 0; index < 3; ++index) {
-            out_cube_a_position[index] = intersection_cube_position[index] + this->m_intersection_zero_as_cube_a_index[index];
-        }
+    void to_cube_a_voxel_position(float* out_cube_a_position, float const* intersection_cube_position) const {
```
sveinung-r: Why does this take raw pointers and not a vector?

In fact, perhaps we should consider std::array for these?

tests/gtest/datahandle_attribute_test.cpp
```diff
@@ -332,4 +332,39 @@ TEST_F(DatahandleAttributeTest, Attribute_Double_Data_Outside_Intersection_Verti
                 testing::ThrowsMessage<std::runtime_error>(testing::HasSubstr("Vertical window is out of vertical bounds")));
 }
 
+TEST_F(DatahandleAttributeTest, Attribute_Double_Different_Number_Of_Samples_To_Border_Per_Dimension) {
```
sveinung-r, line 335: Some ascii art here please.

sveinung-r (changes requested):
See comments

achaikou (commented):
Thnx for the review! :cat:

tests/gtest/datahandle_attribute_test.cpp
sveinung-r: Okay, no need to draw, but I would still like to see a better description of what is tested for. I guess the issue is that I have no clue what you mean by border. Is border the outer cube?

Feel free to demonstrate concepts in 1 or 2 dimensions. I will be able to extend the idea to 3 or more dimensions.

achaikou: You find the comment I've added not to be enough for understanding the test point?

(And yes, it is border of outer cube).

tests/gtest/datahandle_attribute_test.cpp
sveinung-r: Still missing a comment here

achaikou: I added "lowest sample belonging to intersection is 20" that confirms your assumption.

Given that point of the test is to check that when you request data outside of the intersection in vertical plane, you get an error, I am not sure what more besides this comment and test name to add here.

sveinung-r: I don't see that comment anywhere

achaikou: https://github.com/equinor/vds-slice/blob/73c191341ce35a7c36a98eb7e605f6f52be97bce/tests/gtest/datahandle_attribute_test.cpp#L332

sveinung-r: Weird, I don't see it in any of the fixups, but I do see it in "files changed"

achaikou: First fixup, second change from the bottom.

sveinung-r (approved):

PR #273, opened 2024-07-10
Add trusted proxies option

sveinung-r:
Add an option to set a list of trusted proxies when starting the server. This option will determine which IP is considered to be the source IP of a request when headers that contain alternative IP adresses are set (if the proxy is trusted the IP the request is forwarded for will be considered rthe source).

As far as we know the only part of our system
that currently will be impacted by this is the logging.

### DIFF
--- a/cmd/query/main.go
+++ b/cmd/query/main.go
@@ -26,6 +26,7 @@ type opts struct {
 	cacheSize       uint64
 	metrics         bool
 	metricsPort     uint32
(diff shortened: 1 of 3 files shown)


achaikou (commented):
I'd consider you to be the network guy from now on :smile: 

After fixes lets deploy on playground and see how it works first.

achaikou (approved):

PR #275, opened 2024-07-22
Index to annotation in cube intersection gtests

achaikou:
Changes the way slice/fence/attributes gtests perform verifications and make all 3 to use the same class.
(metadata tests also can be unified with them, in theory, but it is separate enough issue I didn't feel the compelling need to do)

### DIFF
--- a/tests/gtest/test_utils.hpp
+++ b/tests/gtest/test_utils.hpp
@@ -53,5 +53,15 @@ public:
     DoubleDataHandle double_reverse_datahandle;
     DoubleDataHandle double_different_size;
 
+    /**
+     * @brief Checks that retrieved value and expected iline, xline and sample match as:
+     *
+     * Each value in the test files has the positional information encoded in little-endian.
+     * Byte 0: Sample position
+     * Byte 1: CrossLine position
+     * Byte 2: InLine position
+     */
+    void check_value(int value, int expected_iline, int expected_xline, int expected_sample);
+
 };
 #endif // TEST_UTILS_H
--- a/tests/gtest/test_utils.hpp
+++ b/tests/gtest/test_utils.hpp
@@ -63,5 +63,75 @@ public:
      */
     void check_value(int value, int expected_iline, int expected_xline, int expected_sample);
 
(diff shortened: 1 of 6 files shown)


tests/gtest/test_utils.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include "ctypes.h"
+#include <iostream>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+const std::string CREDENTIALS = "";
+static constexpr float fill = -999.25;
+
+class DatahandleCubeIntersectionTest : public ::testing::Test {
```
sveinung-r, line 14: Should we make separate classes for each type of handle? I think that would make the tests more readable.

I think all the methods you define on this class in later commits could be free functions.

sveinung-r (changes requested):

tests/gtest/test_utils.hpp
achaikou: How would defining separate class for SingleDatahandleCubeIntersectionTest and DoubleReverseDatahandleCubeIntersectionTest and DoubleDifferentSizeDatahandleCubeIntersectionTest make tests more readable?

The tests would differ only in class name, and I would have to either repeat the same information in every class (which I don't like) or create a parent class with all this information (which we established you don't like).

Now one has to look at the datahandle name to figure out what it represents, otherwise you'll have to look at the class name. I'd say variable name would contain less information, thus easier to process than class name.
So I see zero benefit. For now defining class by test category is totally fine. 

Regarding free functions: then I will have to either rename these functions or add namespaces, as for the new class for tests based on 10 samples testdata the logic would be the same - I would also check attributes, slices and fences.
Encapsulation of logic in the class is more straightforward to me.

sveinung-r: I don't have anything against inheritance in general, only when it ties together code that should not necessarily be tied together.

If you prefer to keep it as is I'm okay with that.

sveinung-r (approved):

PR #284, opened 2024-10-17
Use new openssl in open-vds

achaikou:
Looks like the reason for slowness I've been investigating is openssl version.

Test results on slice retrieval for 500_300_2000 file with 2.5 minutes wait time between requests (which is closer to standard user pattern than hammering requests) :

Before:
`   ✓ request_time_Manual_execution...: avg=1.79s     min=1.67s    med=1.77s    max=1.92s    p(95)=1.92s    p(99)=1.92s    count=8`
After:
`   ✓ request_time_Manual_execution...: avg=1.07s    min=912.93ms med=1.06s    max=1.41s    p(95)=1.31s    p(99)=1.39s    count=8`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,13 +11,18 @@ RUN apk --no-cache add \
     util-linux-dev \
     perl \
     libuv-dev \
-    zlib-dev
+    zlib-dev \
+    openssl-dev
 
 ARG OPENVDS_VERSION=3.4.1
 WORKDIR /
 RUN git clone --depth 1 --branch ${OPENVDS_VERSION} https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds.git
 WORKDIR /open-vds
 
+# for performance reasons open-vds should be build and run with new openssl
+# libssl is installed on alpine by default and would be used during runtime
+
+# it is unclear if curl version makes a difference and which one is better
 ARG BUILD_TYPE=Release
 RUN cmake -S . \
     -B build \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -25,11 +32,17 @@ RUN cmake -S . \
     -DBUILD_EXAMPLES=OFF \
     -DBUILD_TESTS=OFF \
     -DBUILD_DOCS=OFF \
-    -DDISABLE_AWS_IOMANAGER=ON \
+    -DDISABLE_STRICT_WARNINGS=OFF \
     -DDISABLE_AZURESDKFORCPP_IOMANAGER=OFF \
+    -DDISABLE_AWS_IOMANAGER=ON \
```
yngve793, line 37: It is nitpicking. Can we reduce the change by moving this line to the top.

yngve793 (commented):
How do I run the performance test?

Dockerfile
achaikou: It was done on purpose.
Before the thing we are interested in (azure) was hidden in between stuff that we turned off.
Now I moved it up so that first we
- turn off general build stuff not interesting for us
- turn on Azure
- turn off other IO we don't care about
- turn on/off build of libraries that we need

achaikou:
1. To run performance tests you can go to Actions. There you have two workflows:
- performance tests. Standard suite that would run against whatever variables are set in the environment secrets. Needed ones are mentioned in the workflow.
- performance (manual) - where you can run single script and provide all the parameters yourself.

You can't run the file I introduce in this PR as it doesn't belong to any branch yet.
There are almost no runs there as we do not run tests as any part of our workflow. I usually run them from my own fork: turn on test-large environment and run whatever branch I am developing.

2. The other option is to run tests locally. Refer to the Readme "Performance test suite" section.

Note that currently test-large already has updated code.

yngve793 (approved):
Test script "script-random-inlineslice-wait.js" tested successfully. 
No further comments.

PR #285, opened 2024-10-29
Openvds azure manager thread count

achaikou:
Master failing on memory tests: tried to investigate why debug build fails myself, also failed, so submitted [issue](https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds/-/issues/251) to openvds. They don't have resources to work on it, so reverting.


ajaust (approved):
Looks good to me. 🙂

PR #286, opened 2024-11-01
Monitor openvds with snyk

achaikou:
What we discussed - we do not monitor openvds and we probably should, to not miss some critical issue.

I am not sure if these would somehow be repeated by snyk every couple of days (or similar), or if we would want to run the workflow on schedule.
```
  schedule:
    - cron: "05 05 * * MON"
```
We could see how it goes after some time.

### DIFF
--- a/.github/workflows/snyk.yaml
+++ b/.github/workflows/snyk.yaml
@@ -5,6 +5,8 @@ on:
   workflow_dispatch:
 
 jobs:
+  # The following secrets are required:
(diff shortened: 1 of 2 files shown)


ajaust (approved):
Looks good to me. 

We should not need a cronjob here. Snyk will rescan uploaded artifacts every `X` (unsure about value of`X`) hours. I think they rescan everything about 1-2 times a day. As long as the artifact has not changed there is no need to upload a new version to Snyk again.

PR #288, opened 2024-11-14
Use ARM in Radix setups

achaikou:
I will run additional comprehensive tests on test-large once it is merged.
Tests on playground (with less powerful machines) showed that performance difference is inside error margin.

I am kinda worried that ARM actually might have worse performance in our case, but can't find anything conclusive in the web.

Also an option is to switch only test-large environment to arm, but I am not fond of having production and test-large one to be different.

Github doesn't provide standard arm machines yet, so we can't switch tests to them.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 ARG OPENVDS_IMAGE=openvds
-ARG ONESEISMIC_BASEIMAGE=golang:1.22-alpine
+ARG ONESEISMIC_BASEIMAGE=docker.io/library/golang:1.22-alpine
 FROM ${ONESEISMIC_BASEIMAGE} as openvds
 RUN apk --no-cache add \
     git \
--- a/radixconfig_playground.yaml
+++ b/radixconfig_playground.yaml
@@ -5,6 +5,8 @@ kind: RadixApplication
 metadata:
   name: oneseismictest
 spec:
+  build:
+    useBuildKit: true
   environments:
     - name: test
       build:
@@ -12,6 +14,8 @@ spec:
   components:
     - name: server
       dockerfileName: Dockerfile
+      runtime:
+        architecture: arm64
       ports:
         - name: http
           port: 8080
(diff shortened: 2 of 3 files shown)


radixconfig_playground.yaml
```diff
@@ -33,6 +37,7 @@ spec:
             ONESEISMIC_API_TRUSTED_PROXIES: ""
             ONESEISMIC_API_BLOCKED_IPS: ""
             ONESEISMIC_API_BLOCKED_USER_AGENTS: ""
+            OPENVDS_AZURE_IOMANAGER_THREAD_COUNT: ""
```
ajaust, line 40: Is this change ARM specific?

radixconfig_playground.yaml
```diff
@@ -43,8 +48,8 @@ spec:
           monitoring: true
           resources:
             requests:
-              memory: "500Mi"
-              cpu: "200m"
+              memory: "8192Mi"
+              cpu: "2000m"
```
ajaust, line 52: Why do we request mpre more memory and CPU than x86 on playground, but not on production? Is this due to the (large) tests?

Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG OPENVDS_IMAGE=openvds
-ARG ONESEISMIC_BASEIMAGE=golang:1.22-alpine
+ARG ONESEISMIC_BASEIMAGE=docker.io/library/golang:1.22-alpine
```
ajaust, line 2: Why is this change needed? Does docker pull the image from another registry if we do not specify `docker.io/library`?

ajaust (approved):
Looks good overall. I left some small comments so I better understand what is going on. 🙂

radixconfig_playground.yaml
achaikou: No, it is not.

I look at playground config like something third-priority now. It is useful sometimes, most of the time it isn't.

So last time I updated the variables I for some reason didn't update playground file. With this new change it was of interest, as I wanted to test arm performance and you can't do that without changing the config and config is the same for all environments.
So I decided to add the missing parts, but didn't want to create a separate commit. Probably should have had.

radixconfig_playground.yaml
achaikou: These now are the same configs as in production/test-large.

As I was comparing performance I wanted specs to be as similar as possible now. Decided to include in the commit as it makes more sense to have expected characteristics whenever we need to recreate playground environment.

PR #289, opened 2024-11-28
Update performance suites

achaikou:
PR changes performance suits to correspond better with user experience in aspects like data shape and user patterns.

(Look at my fork if you want to see runs of suits).

(It can be divided into smaller PRs if not reviewed by January :cat:  )


tests/performance/helpers/horizon-helpers.js
```diff
@@ -58,31 +58,172 @@ export function retrieveMetadataForHorizonRequest(metadata) {
...
-  let rowArray = new Array(columns).fill(depth);
-  let values = new Array(rows).fill(rowArray);
+function createFlatIncreasingHorizon(
+  rows, columns, depthMin, depthMax
+) {
+  function makeNoise(weight) {
+    return (Math.random() - 0.5) * weight;
+  }
+
+  let points = createEquidistantPoints(depthMin, depthMax, rows + columns - 1, false)
```
yngve793, line 71: Are we only creating points along two edges of the surface?

tests/performance/performance.py
```diff
@@ -1,16 +1,45 @@
 import os
+import re
```
yngve793, line 2: Really appreciate that the scripts can be run locally. I assume that is beneficial for development and testing.

README.md
```diff
@@ -148,13 +148,13 @@ To locally run a file `<chosen_script>.js` from the suite:
     ```
     export PYTHONPATH=<path-to-the-repo>/oneseismic-api/tests:$PYTHONPATH
     ```
-4. Set all environment variables
+4. Set all required environment variables
```
yngve793, line 151: Running "script-constant-fence.js " using the test environment from `tests/performance/requirements-dev.txt`give the following error:

└─(12:35:17,   18 files, 192b)─> python tests/performance/performance.py tests/performance/script-constant-fence.js 
Traceback (most recent call last):
  File "tests/performance/performance.py", line 6, in <module>
    from utils.cloud import generate_account_signature
ModuleNotFoundError: No module named 'utils'

yngve793 (approved):
The PR uses Github workflows for the performance tests. Running the tests in the cloud makes the results independent of the developers personal device. It is slightly more unclear how predictable the specifications and load of the available resources are.  

Overall the code looks good when reading through it. The PR builds on top of existing performance test that I do not have experience with.  The tests seams to work as intended and the structure of the code makes it easy to read. 

Recommend approval with low confidence.

README.md
achaikou: That most likely means that you didn't add tests directory to PYTHONPATH, as is mentioned in the readme.

In the tests directory we have a directory called "utils" which is responsible for sas creation. It is used for performance, memory and e2e tests.
Because of the pythons way to deal with importing code from non-current python file (or the way I ~don't~ understand it) , it had to become a common module and python must be told where to search for this module.

yngve793 (approved):
Good, rebase and merge.

PR #291, opened 2025-01-15
Use timezone aware object for UTC

yngve793:
Fix DeprecationWarning.

DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    expiry = datetime.datetime.utcnow() + datetime.timedelta(seconds=lifespan)

### DIFF
--- a/tests/e2e/Dockerfile
+++ b/tests/e2e/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-slim AS e2etest
+FROM python:3.12-slim AS e2etest
 
 COPY ./tests/e2e/requirements-dev.txt /tests/e2e/requirements-dev.txt
 RUN python -m pip install --upgrade pip
--- a/tests/utils/cloud.py
+++ b/tests/utils/cloud.py
@@ -13,7 +13,7 @@ def generate_account_signature(
     """
     Create account signature for azure request
     """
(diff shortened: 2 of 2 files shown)


achaikou:
`AttributeError: module 'datetime' has no attribute 'UTC'`
?
https://github.com/achaikou/oneseismic-api/actions/runs/13205562221/job/36867813490#step:5:191

ajaust:
Just FYI:
- The E2E tests run on [Python 3.9](https://github.com/equinor/oneseismic-api/blob/d3c7e04121c449a7810029ec87878d3ed1e84cf2/tests/e2e/Dockerfile#L1).
- `datetime.UTC` was introduced in [Python 3.11](https://docs.python.org/3/library/datetime.html#datetime.UTC).
- `utcnow()` was deprecated in [Python 3.12](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow).

achaikou:
Should we maybe instead then switch running tests to python >3.11?
3.9 will be EOL this year, so we will have to upgrade anyway.

achaikou (approved):
Nice! Rebase and merge :sun_with_face:

yngve793 (commented):
Force pushed update to Python 3.12

PR #292, opened 2025-01-22
Upgrade openvds to 3.4.5

achaikou:
1. Fixes fmt library issue that breaks compilation
2. Allows us to remove some build dependencies.

Memory tests supressions had to be updated, but hopefully they will be removed soon due to fix in the source code.


yngve793 (approved):
Nice to have a fix-up. 
Builds and runs without problems.

PR #293, opened 2025-02-13
Fix casing do not match warnings

yngve793:
Building the docker image give the following warning:
- FromAsCasing: 'as' and 'FROM' keywords' casing do not match

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG OPENVDS_IMAGE=openvds
 ARG ONESEISMIC_BASEIMAGE=docker.io/library/golang:1.22-alpine
-FROM ${ONESEISMIC_BASEIMAGE} as openvds
+FROM ${ONESEISMIC_BASEIMAGE} AS openvds
 RUN apk --no-cache add \
     git \
     g++ \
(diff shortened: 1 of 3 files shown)


achaikou (approved):
Builds fine :+1: 
Do you want to fix other dockerfiles, like memory one, too?

yngve793 (commented):
Applied similar fix to tests/memory/Dockerfile

Please have a look

achaikou (commented):
If you are fixing legacy ENV key-value assignment, would it be also useful to fix it in the tests/performance/Dockerfile? :innocent:

yngve793 (commented):
Performance dockerfile fixed. Not tested. 

All three files give warning of the type:
- UndefinedVar: Usage of undefined variable '$PYTHONPATH' (line 28)

As far as I can tell the variable is empty or not set, but I presume that it is added for a reason.

achaikou (approved):
Looks nice!

I thought PYTHONPATH might be set from before, as it might happen on some systems, but maybe not in docker.
So probably we can just set it, without appending, if you want to get rid of this warning too. 
https://stackoverflow.com/questions/49631146/how-do-you-add-a-path-to-pythonpath-in-a-dockerfile#comment139242041_49631407

PR #304, opened 2026-04-14
Upgrade to Go 1.26

achaikou:
- Radix build is not tested, but do not expect any problems.
- To discuss if we want to pin to commit in any of the github actions.


ajaust (approved):
Looks good. :)

PR #305, opened 2026-07-27
Repository maintenance

ajaust:
- Bump Go dependencies to latest version
- Added a dependabot configuration to have more fine-grained control over updates. It covers Go, Python, Docker and GitHub Actions.
- Updated GitHub Actions to newest version + pinned to commit hash of release.
- Pinned Docker containers to current digest (=container hash)

I was curious if we could move OpenVDS to the newest release, i.e. [3.4.9](https://community.opengroup.org/osdu/platform/domain-data-mgmt-services/seismic/open-vds/-/releases/3.4.9), but the build process failed due to some linking issues w.r.t. to OpenVDS cURL handler, see [GitHub Action logs](https://github.com/ajaust/oneseismic-api/actions/runs/30265902348/job/89976376339). I decided to not dig into this for now.


.github/dependabot.yml
```diff
@@ -0,0 +1,54 @@
...
+          - "*"
+    cooldown:
+      default-days: 7
+
+  - package-ecosystem: "go"
+    directory: "/"
+    schedule:
+      interval: "quarterly"
+    allow:
+      - dependency-type: "direct"
```
achaikou, line 20: (I secretly hate dependabot because I can't understand its PRs, but lets try it 😄  )

Are you sure it should be "direct" and not "all"?
As usually with dependabot I completely fail to understand what it is doing and why it is doing it, so I can't find consequences of this choice.

.github/dependabot.yml
```diff
@@ -0,0 +1,54 @@
...
+      interval: "quarterly"
+    groups:
+      docker:
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 7
+
+  - package-ecosystem: "pip"
+    directories:
```
achaikou, line 41: Did you exclude `tests/e2e` on purpose?
If not, should pip and docker use [recursive](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files) setup?

(doesn't matter much though as there was no need to pin the dependencies then)

achaikou (approved):
Yay! Our security master has arrived and fixed everything! 🦺 😎 

---

Ah, 1.27 is to be released in August. We'll upgrade some other time then. Would need to deprecate python 11 soon too.

Regarding update for openvds we also discussed it [a bit](https://github.com/equinor/oneseismic-api/pull/302#discussion_r2671123908) beginning of the year.
Don't remember anymore if this is the error we had before or a new one 😄 

Maybe later this year we could set aside time to try digging deeper into all those updates.

.github/dependabot.yml
ajaust: I appreciate your openness to be more annoyed by dependabot. If it turns out to be horrible, I will take the blame. We can drop/adapt the config if we don't like it. The configuration was my attempt to limit the type/amount of PRs dependabot creates, but now I see that dependabot was quite tame in this repository anyway (compared to other repos). 

I chose `direct` in an effort to limit the PRs that update dependencies. `direct` should only update dependencies that we have explicitly defined. It should ignore updates of transitive/indirect dependencies. This is at least how I understand [the docs on `dependency-type`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#dependency-type-allow).

Security updates, also in transitive dependencies, should be unaffected and always trigger PRs.

.github/dependabot.yml
ajaust: Good catch. This was my oversight. Looking into this now, I also see that it did not cover the `requirements-dev.txt` in `testdata/`. Let's try the recursive approach (added as fixup) and see how this works.

ajaust (commented):
Thanks for the review. I addressed your comments. 🙂

.github/dependabot.yml
achaikou: From what I understand, `quarterly` + `*groups` should already limit the PR to just one. Maybe I think too highly of it though.

But as dependabot doesn't write what rule created the PR and the command that it used, I have no clue if "direct" or "all" would be the same thing we are doing manually or not.

And no need to take any blame for anything, lets see how it works and how not-in-this-repo-style commit messages it would generate :)

PR is still very much approved :)

PR #306, opened 2026-07-28
Fix dependabot config

ajaust:
It turns out that the correct ecosystem name for Go is `gomod` instead of `go`. The PR should resolve [this failing job](https://github.com/equinor/oneseismic-api/runs/90247398980).


achaikou (approved):
FYI: I did yesterday ask AI whether contents of this file is sensible 😄 

It was pretty happy and kept calling the package-system "go" 😄

